PR #18, opened 2017-07-13
ORC IO

### LINKED ISSUE #8: OSM => VectorTiles :: (2) Reading `RDD[Element]` from an ORC file
ORC files lend themselves to being read in parallel. Via AWS Athena, they can apparently been queried by SQL. This give us a window into the data via Spark, to parse the entire Earth in parallel, as previously couldn't be done with our XML approach.

**Questions:**

- What's the simplest setup to begin querying an ORC file?
- How do we parse the results?
- What is Spark's role? Can we get an `RDD` easily?

fosskers:
### TODO

- [x] `fromLocalORC`
- ~~`fromRemoteORC` ?~~ (`fromLocalORC` is now `fromORC`)
- [x] Testing
- [x] Docs

Closes #8 .

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -18,6 +18,7 @@ libraryDependencies ++= Seq(
   "io.dylemma"                  %% "xml-spac"              % "0.3-SNAPSHOT",  // Published locally
   "org.apache.spark"            %% "spark-core"            % "2.2.0",  // TODO add "provided" later
   "org.apache.spark"            %% "spark-hive"            % "2.2.0",
+  "org.apache.hadoop"           %  "hadoop-aws"            % "2.8.1",
   "org.locationtech.geotrellis" %% "geotrellis-spark"      % "1.1.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-util"       % "1.1.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-vector"     % "1.1.0-SNAPSHOT",
--- a/build.sbt
+++ b/build.sbt
@@ -16,7 +16,8 @@ scalacOptions := Seq(
 libraryDependencies ++= Seq(
   "com.azavea"                  %% "scaliper"              % "0.5.0-SNAPSHOT" % "test",
   "io.dylemma"                  %% "xml-spac"              % "0.3-SNAPSHOT",  // Published locally
-  "org.apache.spark"            %% "spark-core"            % "2.1.0",  // TODO add "provided" later
+  "org.apache.spark"            %% "spark-core"            % "2.2.0",  // TODO add "provided" later
+  "org.apache.spark"            %% "spark-hive"            % "2.2.0",
   "org.locationtech.geotrellis" %% "geotrellis-spark"      % "1.1.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-util"       % "1.1.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-vector"     % "1.1.0-SNAPSHOT",
--- a/build.sbt
+++ b/build.sbt
@@ -17,7 +17,7 @@ libraryDependencies ++= Seq(
   "com.azavea"                  %% "scaliper"              % "0.5.0-SNAPSHOT" % "test",
   "io.dylemma"                  %% "xml-spac"              % "0.3-SNAPSHOT",  // Published locally
   "org.apache.spark"            %% "spark-core"            % "2.2.0",  // TODO add "provided" later
-  "org.apache.spark"            %% "spark-hive"            % "2.2.0",
+  "org.apache.spark"            %% "spark-sql"             % "2.2.0",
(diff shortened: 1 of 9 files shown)


mojodna:
`fromRemoteORC` is probably `fromLocalORC` with URIs that point to remote sources (e.g. `s3a://...`)

fosskers:
Probably! I haven't looked into the remote case at all yet. I was assuming there would be some AWS magic. I remember the word "Athena"?

mojodna:
Nah, in this case you're just looking at Hadoop / HDFS / other filesystem "magic."

Athena is AWS's hosted Presto service.

fosskers:
This is very good news.

fosskers:
`fromLocalORC` timings in seconds, counting Elements on a 178mb file:

Nodes: 32.3s to count 15.4mil nodes
Ways: 18.7s to count 1.49mil ways
Rels: 11.5s to count 86k relations

Interestingly, my `streaming-osm` lib can count that same number of nodes in ~30s with a single core.
Spark overhead must be pretty heavy here, but the good sign is that our time isn't linear/quadratic. i.e. Counting the rels might be "slow", but to count 2 orders of magnitude more nodes doesn't take 100 times as long.

This _should_ speed up nicely with a big EMR cluster.

lossyrob:
> Interestingly, my `streaming-osm` lib can count that same number of nodes in ~30s with a single core.

How many cores is spark using? Making sure Spark is configured to utilize all cores and memory and increasing the partition count could potentially speed things up, even in the local case.

fosskers:
I'm using `local[*]` and I see it grinding away with all 4 cores on my machine. I always forget about partition count, so tweaking that might speed things up.

`fromORC` is ass-slow when dealing with large files on S3, but that's excected. Should magically be faster on EMR because of its special connection to S3.

build.sbt
```diff
@@ -16,7 +16,9 @@ scalacOptions := Seq(
 libraryDependencies ++= Seq(
   "com.azavea"                  %% "scaliper"              % "0.5.0-SNAPSHOT" % "test",
   "io.dylemma"                  %% "xml-spac"              % "0.3-SNAPSHOT",  // Published locally
-  "org.apache.spark"            %% "spark-core"            % "2.1.0",  // TODO add "provided" later
+  "org.apache.spark"            %% "spark-core"            % "2.2.0",  // TODO add "provided" later
```
lossyrob: You can use

```
fullClasspath in Runtime := (fullClasspath in (Compile, run)).value
```

in the build file to allow for these to be marked as `provided` while still allowing `sbt run`

fosskers: `sbt` is deeply magical to me, and I add `provided` on Spark deps mostly for cargo-cult reasons. What problem does your suggested addition solve? (honest question)

src/main/scala/vectorpipe/osm/package.scala
```diff
@@ -27,48 +27,125 @@ package object osm {
...
-    *   - Graphs of [[Relation]]s will be broken into spanning [[Tree]]s.
-    *   - It doesn't make sense to represent non-multipolygon Relations as
-    *     GeoTrellis `Geometry`s, so Relation metadata is disseminated
-    *     across its child members. Otherwise, Relations are "dropped"
-    *     from the output.
+  /** Given a path to an Apache ORC file containing OSM data, read out RDDs of each Element type.
+    * If you want to read a file from S3, you must call [[vectorpipe.useS3]] first
+    * to properly configure Hadoop to read your S3 credentials.
     */
+  def fromORC(path: String)(implicit hc: HiveContext): Either[String, (RDD[Node], RDD[Way], RDD[Relation])] = {
```
fosskers: Maybe `HiveContext` should be created *inside* this function, so that the user doesn't have to worry about it at all. It would save them an explicit dependency and an import in their main file.

build.sbt
lossyrob: Well, it should solve the fact that when you do `sbt assembly`, to run it against some spark submit (e.g. EMR), that it doesn't bake in the Spark and Hadoop dependencies, because those are provided by `spark-submit`.

I'm testing on the project I needed that for now, and it looks like the assembly is actually still baking in spark and hadoop; so actually I'll need to find a better solution.

The thing that this is trying to solve is, when these dependencies are marked "provided", `sbt run` will error with ClassNotFound's. This attempts to put the correct files on the class path in the "run" situation, so that doesn't happen.

fosskers: Much clearer, thanks. :+1:

lossyrob: Working version of this:

```
run in Compile := Defaults.runTask(fullClasspath in Compile, mainClass in (Compile, run), runner in (Compile, run)).evaluated
runMain in Compile := Defaults.runMainTask(fullClasspath in Compile, runner in(Compile, run)).evaluated
```

fosskers: Hot damn, thanks!

fosskers: Unfortunately this breaks Microsite building.

PR #19, opened 2017-09-05
Skip and log clipping failures

fosskers:
### TODO

- [x] `toGrid` expects `clip` function to return an ~~`Either`~~ `Option`
- [x] Built-in clipping functions yield `Either`
- ~~Log any `Left`s given from clipping~~ (logging is now framework agnostic)
- [x] Make logging "framework agnostic" and provide sensible defaults 

### Motivation

We've decided to assume that exotic, non-cooperative geometries will always exist within OSM data. Rather than crash the whole ingest when one of these is encountered, this PR allows for these to be skipped over and logged. It should be possible to log any `Left` values encountered without incurring too much invasive `IO` - hopefully transfer of the failure string to the driver or a write to S3 will be enough.

**Question:** If we naively print a logging statement via each executor, will those be aggregated into the EMR log file already written to S3?

**Question:** Where to print the errors in the local case? It would be nice if a single solution took care of both the local and EMR cases.

Ping @lossyrob

### DIFF
--- a/src/main/scala/vectorpipe/Clip.scala
+++ b/src/main/scala/vectorpipe/Clip.scala
@@ -74,7 +74,7 @@ object Clip {
     centre.distanceToSegment(p1, p2) <= radius
 
   /** Naively clips Features to fit the given Extent. */
-  def byExtent[G <: Geometry, D](extent: Extent, f: Feature[G, D]): Either[String, Feature[Geometry, D]] = {
+  def byExtent[G <: Geometry, D](extent: Extent, f: Feature[G, D]): Option[Feature[Geometry, D]] = {
     val exPoly: Polygon = extent.toPolygon
 
     val clipped: Try[Geometry] = f.geom match {
@@ -82,10 +82,7 @@ object Clip {
       case _ => Try(f.geom.intersection(exPoly).toGeometry.get)
     }
 
-    clipped match {
-      case Failure(_) => Left(s"${f.data}: ${f.geom.reproject(WebMercator, LatLng).toGeoJson}")
-      case Success(g) => Right(Feature(g, f.data))
-    }
+    clipped.toOption.map(g => Feature(g, f.data))
   }
 
   /** Clips Features to a 3x3 grid surrounding the current Tile.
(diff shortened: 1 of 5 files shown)


lossyrob:
> If we naively print a logging statement via each executor, will those be aggregated into the EMR log file already written to S3?

We'd want to have a pluggable way to log these bad geometries. In the 'real system', I am imagining pushing these to something like an SQS queue. So naively logging to output won't do the trick, though might be an acceptable default for things like local tests.

> Where to print the errors in the local case? It would be nice if a single solution took care of both the local and EMR cases.

I don't think a single solution will cover both cases, but a single architecture could let users plug in their own logger/aggregator that has a sensible local default.

fosskers:
> I am imagining pushing these to something like an SQS queue. 

Neat, so it could be an additional parameter to `toGrid` that provides the equivalent of:
```scala
logIt: String => IO[Unit]
```
be that a standalone function, some logging class, etc.

fosskers:
Caveat to the above: the contents of `logIt` need to be serializable, since the log calls are made within an RDD `.map` step.

lossyrob:
Can we get more info than a String? I'm imagining a case class that contains the offending OSM elements, or other information that's specific and complete to dealing with the bad data.

fosskers:
The current try-catch logic around the topology exceptions [*kinda* does this](https://github.com/geotrellis/vectorpipe/pull/19/files#diff-edce1f6357aabcc0af9e4a27f204de84R86). The clipping errors are caught and logged per-failure by necessity, so a wrapping case class could only wrap the `Feature` that failed. Maybe just return the `Feature` entirely, and demand:
```scala
logIt: Feature[G, D] => IO[Unit]
```
Thoughts?

lossyrob:
If the feature that fails contains enough info to identify the OSM elements that caused the failure, I believe that would be enough.

We probably want the clipping extent as well, so what you end up with in the log of failed things is something like Seq[(Extent, Feature[G, D])].

We could stream that into a queue and then have a backlog of geometries to inspect to run down errors.

lossyrob:
Might actually just need to be `(Extent, Feature[G, D]) => IO[Unit]` to account for the streaming queue update

fosskers:
> If the feature that fails contains enough info to identify the OSM elements that caused the failure, I believe that would be enough.

It would, thanks to our nested `ElementData` tree.

> We probably want the clipping extent as well

Good idea. Luckily that `Extent` is visible to the clipping function, so if the clip fails it can dump back the `Extent` along with the `Feature`.

fosskers:
I can make those changes right after RRP.

fosskers:
Questions/notes for @lossyrob:

- The clipping functions now return in `Option` instead of `Either`, since both the original geom and the `Extent` are already visible to the parent method (`toGrid`)
- A logging function is now expected for `toGrid`, as discussed, which lets us be 100% agnostic about the logging framework
- A few default logging functions were provided:
  - `stdout`: dump the entire geom and its metadata to stdout
    - has no insight into what the `D` is, so it can only print the whole thing
  - `osm.stdout`: dump the entire geom with only its Element ID (not the entire, massive metadata tree)
    - specialized: `D := Tree[ElementData]`
    - less verbose than the first `stdout`
  - `ignore`: don't bother logging errors
- Design question: how to name the logging functions? Right now they just say what they do without further flourish. I could imagine `logStdout`, prefixing them all with `log` so that they appear in a nice order in the Scaladocs. Thoughts?
- Design question: I'm not entirely sure where these logging functions should live. Suggestions:
  - **Current approach:** in the package they belong to (like `osm.stdout`), or at top-level in `VectorPipe` if they apply to any data source
  - in the `Clip` object, since they primarily involving logging clipping failures
  - in some new `vectorpipe.log` package
  - in some new `vectorpipe.Log` object

Regarding the last point: through a lot of head-banging and refactoring, I've made vectorpipe usable with just `import vectorpipe._`. Trying to reduce the number of named symbols (classes/objects) as much as possible was part of that, so I'm wary of a new `Log` object. Thoughts?

fosskers:
A default logging function that uses `log4j` is probably also desirable... although I fear serialization demons, haha.

fosskers:
```scala
/* Shouldn't have to worry about `logger` not being serializable if we call for it every time? */
def log4j[G <: Geometry, D](e: Extent, f: Feature[G, D]): Unit = {
  val msg = ... // get the string to print
  val logger = Logger.rootLogger()

  logger.error(msg)
}
```

lossyrob:
I'd have to work with the API a bit more to be able to help give an opinion on the design questions, but if you're organizing things into objects and it applies only to Clip failure logs, I'd say putting them in Clip makes sense to me at the moment.

If you call the object Logger method `rootLogger()` every time, that should avoid any serialization errors.

PR #21, opened 2017-10-06
De-emphasize AVTs

fosskers:


build.sbt
```diff
@@ -6,7 +6,7 @@ organization := "com.azavea"
 
 organizationName := "Azavea"
 
-scalaVersion := "2.11.11"
+scalaVersion in ThisBuild := "2.11.11"
```
lossyrob: What does this do?

build.sbt
```diff
@@ -15,20 +15,19 @@ scalacOptions := Seq(
...
   "org.locationtech.geotrellis" %% "geotrellis-spark"      % "1.2.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-util"       % "1.2.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-vector"     % "1.2.0-SNAPSHOT",
   "org.locationtech.geotrellis" %% "geotrellis-vectortile" % "1.2.0-SNAPSHOT",
-  "org.scalatest"               %% "scalatest"             % "2.2.0" % "test",
-  "org.typelevel"               %% "cats"                  % "0.9.0",
-  "org.spire-math"              %% "spire"                 % "0.13.0"
+  "org.scalatest"               %% "scalatest"             % "3.0.1" % "test",
+  "org.spire-math"              %% "spire"                 % "0.13.0",
+  "org.typelevel"               %% "cats-core"             % "1.0.0-MF"
```
lossyrob: What is MF? Is this an release?

src/main/scala/vectorpipe/util/Graph.scala
```diff
@@ -1,158 +0,0 @@
...
-  *     ├── 2
-  *     │   └── 4
-  *     │       └── ∅
-  *     └── 3
-  *         └── ∅
-  * }}}
-  *
-  * @constructor Use `Graph.fromEdges` instead of this.
-  */
-class Graph[K: Order, V] private (
```
lossyrob: Seems like a lot of work went into this and some of the other data structures/code. Should this go somewhere for preservation, in case we need it later?

build.sbt
fosskers: Prevents ensime from freaking out when I run `enimeConfig` every once in a while in sbt. You need to rerun it every time you add a new dependency.

build.sbt
fosskers: "Milestone Final" by their wording.

src/main/scala/vectorpipe/util/Graph.scala
fosskers: ScalaZ has a `Tree`, but I haven't seen a `Graph` anywhere in Scalaland. It could be good to donate the code to either ScalaZ or Cats, or both.

build.sbt
fosskers: It's the version they have displayed as official on their [microsite](https://typelevel.org/cats/).

src/main/scala/vectorpipe/util/Graph.scala
lossyrob: Perhaps would be good to create an issue in the Azavea 10% time repo pointing at the code and talking about potential contribution. There might be some folk (or yourself) that would want to take that on as a 10% time project.

build.sbt
lossyrob: :+1:

build.sbt
lossyrob: :+1:

src/main/scala/vectorpipe/util/Graph.scala
fosskers: That's a good way to not forget about it, yeah. Doing so now.

fosskers: https://github.com/azavea/azavea-10-percent-time/issues/154

PR #22, opened 2017-10-10
Add logging to `toFeatures`

fosskers:
### TODO

- [x] Log `Line` fusion failures
- [x] Provide cleaner logging function API

### Motivation

We were silently skipping over failures to fuse `Line`s together in the depths of `ElementToFeature`. This PR adds a customizable way to log those failures, as well as reworks the APIs of the library's primary functions.

### DIFF
--- a/src/main/scala/vectorpipe/VectorPipe.scala
+++ b/src/main/scala/vectorpipe/VectorPipe.scala
@@ -113,12 +113,9 @@ object VectorPipe {
     * @param clip A function which represents a "clipping strategy".
     * @param logError An IO function that will log any clipping failures.
     */
-  def toGrid[G <: Geometry, D](
-    clip: (Extent, Feature[G, D]) => Option[Feature[G, D]],
-    logError: (Extent, Feature[G, D]) => Unit,
-    ld: LayoutDefinition,
-    rdd: RDD[Feature[G, D]]
-  ): RDD[(SpatialKey, Iterable[Feature[G, D]])] = {
+  def toGrid[G <: Geometry, D](ld: LayoutDefinition, rdd: RDD[Feature[G, D]])
+            (clip: (Extent, Feature[G, D]) => Option[Feature[G, D]])
+            (logError: (((Extent, Feature[G, D])) => String) => ((Extent, Feature[G, D])) => Unit): RDD[(SpatialKey, Iterable[Feature[G, D]])] = {
 
     val mt: MapKeyTransform = ld.mapTransform
 
@@ -132,6 +129,11 @@ object VectorPipe {
     /* Associate each Feature with a SpatialKey */
     val grid: RDD[(SpatialKey, Feature[G, D])] = rdd.flatMap(f => byIntersect(mt, f))
 
+    /** A way to render some Geometry that failed to clip. */
+    val errorClipping: ((Extent, Feature[G, D])) => String = { case (e, f) =>
+      s"CLIP FAILURE W/ EXTENT: ${e}\nELEMENT METADATA: ${f.data}\nGEOM: ${f.geom.reproject(WebMercator, LatLng).toGeoJson}"
+    }
+
     grid.groupByKey().map { case (k, iter) =>
       val kExt: Extent = mt(k)
 
@@ -139,7 +141,7 @@ object VectorPipe {
       val clipped: List[Feature[G, D]] = iter.foldLeft(List.empty[Feature[G, D]]) { (acc, g) =>
         clip(kExt, g) match {
           case Some(h) => h :: acc
-          case None => logError(kExt, g); acc /* Sneaky IO to log the error */
+          case None => logError(errorClipping)((kExt, g)); acc /* Sneaky IO to log the error */
         }
       }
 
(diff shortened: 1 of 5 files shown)


fosskers:
The logging function for `toFeature` and for `toGrid` are embarrassingly similar. I'd like to combine them somehow.

src/main/scala/vectorpipe/VectorPipe.scala
```diff
@@ -113,12 +113,9 @@ object VectorPipe {
...
     */
-  def toGrid[G <: Geometry, D](
-    clip: (Extent, Feature[G, D]) => Option[Feature[G, D]],
-    logError: (Extent, Feature[G, D]) => Unit,
-    ld: LayoutDefinition,
-    rdd: RDD[Feature[G, D]]
-  ): RDD[(SpatialKey, Iterable[Feature[G, D]])] = {
+  def toGrid[G <: Geometry, D](ld: LayoutDefinition, rdd: RDD[Feature[G, D]])
+            (clip: (Extent, Feature[G, D]) => Option[Feature[G, D]])
+            (logError: (((Extent, Feature[G, D])) => String) => ((Extent, Feature[G, D])) => Unit): RDD[(SpatialKey, Iterable[Feature[G, D]])] = {
```
fosskers: Multiple arg lists to be more Scala-like.

src/main/scala/vectorpipe/VectorPipe.scala
```diff
@@ -174,26 +176,22 @@ object VectorPipe {
...
-  def log4j[G <: Geometry, D](e: Extent, f: Feature[G, D]): Unit =
-    Logger.getRootLogger().error(logString(e, f))
+  /** Log an error to STDOUT. */
+  def logToStdout[A](f: A => String): A => Unit = { a => println(f(a)) }
+
+  /** Log an error as an ERROR through Spark's default log4j. */
+  def logToLog4j[A](f: A => String): A => Unit = { a => Logger.getRootLogger().error(f(a)) }
+
+  /** Skip over some failure. */
+  def logNothing[A](f: A => String): A => Unit = { _ => Unit }
```
fosskers: Notice both the `f` and `a` are in scope. One's custom logging function could then choose to ignore whatever default `f` had been passed up from the depths of VectorPipe and render the error in whatever way they want.

src/main/scala/vectorpipe/osm/Element.scala
```diff
@@ -12,7 +12,7 @@ sealed trait Element {
   def data: ElementData
 }
 
-object Element {
+private[vectorpipe] object Element {
```
fosskers: Hides it from Scaladocs.

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
```diff
@@ -271,46 +275,42 @@ private[vectorpipe] object ElementToFeature {
...
   }
 
   /**
    * Fuse the head Line in the Vector with the first other Line possible.
    * This borrows [[fuseLines]]'s assumptions.
    */
-  private def fuseOne(
-    v: Vector[OSMLine]
-  ): State[Int, Option[(Line, ElementData, Int, Vector[OSMLine])]] = {
+  private def fuseOne(v: Vector[OSMLine]): Option[(Line, ElementData, Int, Vector[OSMLine])] = {
```
fosskers: Was able to make this pure again. The `State` stuff is handled back up in `fuseLines`.

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
```diff
@@ -271,46 +275,42 @@ private[vectorpipe] object ElementToFeature {
...
     val h = v.head
     val t = v.tail
 
-    // TODO: Use a `while` instead?
+    // TODO: Use a `while` instead? Scala `for` is slow.
```
fosskers: Let's do something about this.

src/main/scala/vectorpipe/VectorPipe.scala
moradology: Minor nitpick: it seems to me that the first argument should be the one that changes least in a given program. My intuition here is that different clipping strategies might make sense while different logging behaviors would be somewhat strange. Consider inverting the argument order

src/main/scala/vectorpipe/VectorPipe.scala
```diff
@@ -132,14 +129,19 @@ object VectorPipe {
...
+
     grid.groupByKey().map { case (k, iter) =>
       val kExt: Extent = mt(k)
 
       /* Clip each geometry in some way */
       val clipped: List[Feature[G, D]] = iter.foldLeft(List.empty[Feature[G, D]]) { (acc, g) =>
         clip(kExt, g) match {
           case Some(h) => h :: acc
-          case None => logError(kExt, g); acc /* Sneaky IO to log the error */
+          case None => logError(errorClipping)((kExt, g)); acc /* Sneaky IO to log the error */
```
moradology: Being sneaky w/r/t the type system is unavoidable, but I think you should break this out and put the `IO` and the return value on separate lines for human-sake

src/main/scala/vectorpipe/VectorPipe.scala
moradology: I like these example implementations

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
moradology: I prefer `cfor`, but saving the import would be nice

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
```diff
@@ -322,16 +322,13 @@ private[vectorpipe] object ElementToFeature {
 
         val (a, b) = t.splitAt(i)
 
-        /* Hand-hold the typechecker */
-        val res: Option[(Line, ElementData, Int, Vector[OSMLine])] =
-          Some((line, h.data, i, a ++ b.tail))
-
         /* Return early */
-        return res.pure[IntState]
+        return Some((line, h.data, i, a ++ b.tail))
```
moradology: Evidently `dotty` will make it so that these extra parens are a thing of the past (at least for function application)

src/main/scala/vectorpipe/osm/package.scala
```diff
@@ -176,7 +177,8 @@ package object osm {
    *     across its child members. Otherwise, Relations are "dropped"
    *     from the output.
    */
-  def toFeatures(nodes: RDD[Node], ways: RDD[Way], relations: RDD[Relation]): RDD[OSMFeature] = {
+  def toFeatures(nodes: RDD[Node], ways: RDD[Way], relations: RDD[Relation])
+                (logError: (Feature[Line, ElementData] => String) => Feature[Line, ElementData] => Unit): RDD[OSMFeature] = {
```
moradology: Consider the same point about order of argument lists from above

src/main/scala/vectorpipe/VectorPipe.scala
fosskers: > the first argument should be the one that changes least in a given program

I follow that philosophy as well. So really the functions should be first, except "functions last" seems to be a Scala thing. I'm caught on what to do.

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
fosskers: `cfor == while` in my mind so I don't mind how we actually go about it.

src/main/scala/vectorpipe/VectorPipe.scala
moradology: I didn't realize there was best practice around this in the community. So much the worse for scala's standards - I think you should follow your heart and put functions before simpler bits of the ontology

src/main/scala/vectorpipe/VectorPipe.scala
fosskers: Fair, fair.

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
fosskers: In this case I actually want the quadruplet to be visible.

src/main/scala/vectorpipe/osm/package.scala
fosskers: Since there's only one tacked-on function here, I'd lean on Scala convention. Someone might want to:
```scala
toFeatures(nodes, ways, rels) { f => { a => ??? } }
```

src/main/scala/vectorpipe/VectorPipe.scala
lossyrob: https://docs.scala-lang.org/style/declarations.html#multiple-parameter-lists

Multiple args aren't necessarily more scala-like, especially if there are more than one.

The Scala-like separation of a function and the rest of the param list really makes sense when there is usually a lambda being passed in, so that you can get a syntax like this:

```scala
doAThing(gonnaDoItToThis) { x => // the thing that does the thing }
```

This breaks when you have multiple functions that are really just params to the function, and are auxiliary to the main point of the function.

So my 2cents is to remove the multiple params here.

fosskers: It's really doubtful that somebody would implement their custom `clip` as a lambda to `toGrid`. Likewise for the logging functions, so those are used in two places.

fosskers: Although there is a point from the document that @lossyrob just linked:

> For complex DSLs, or with type-names that are long, it can be difficult to fit the entire signature on one line. 

which is true for this case. The code and scaladocs render a bit prettier with multiple arg lists.

PR #23, opened 2017-10-11
Reading Planet History

fosskers:
Branched off #22 

See https://github.com/mojodna/osm2orc/issues/7

PR #25, opened 2017-10-13
Use java.time.instant for timestamps

moradology:
I wasn't able to compile this due to the current dependency upon snapshotted GT, but I think this is what we'll want for timestamp conversions


fosskers:
The ORC code would have to change as well. Notes to self: a small test would have to be performed on some ORC file to guarantee that reflection doesn't freak out with this change.

It's frustrating because compilation isn't a guarantee when it comes to the ORC reading behaviour.

fosskers:
Superseded by #23

PR #26, opened 2017-10-16
Add function for associating ways/nodes

moradology:
This PR adds a method which groups nodes/ways in preparation for ultimately producing a snapshot of each `Way` throughout history


fosskers:
Superseded by #23

PR #27, opened 2017-10-17
Remove scaliper from vectorpipe.

lossyrob:


fosskers:
This has already been removed on `master`, it looks like you forked off an older revision.

PR #35, opened 2017-11-02
Mark hadoop-aws as provided

moradology:
This change will allow our jar to be included in EMR without trouble. This PR doesn't address the `useS3` function which will work for apache's `hadoop-aws` but not for Amazon's proprietary fork of `hadoop-aws`.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -20,10 +20,7 @@ val common = Seq(
 
   libraryDependencies ++= Seq(
     "io.dylemma"                  %% "xml-spac"              % "0.3",
-    "org.apache.hadoop"           %  "hadoop-aws"            % "2.8.1"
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -20,10 +20,7 @@ val common = Seq(
 
   libraryDependencies ++= Seq(
     "io.dylemma"                  %% "xml-spac"              % "0.3",
-    "org.apache.hadoop"           %  "hadoop-aws"            % "2.8.1"
-      exclude("javax.servlet",     "servlet-api")
-      exclude("javax.servlet.jsp", "jsp-api")
-      exclude("org.mortbay.jetty", "servlet-api"),
```
fosskers: These were explicit excludes to get Spark UI to work locally. We should test if it still works.

PR #36, opened 2017-11-07
Fix line order bug

fosskers:
Previously I wasn't respecting the original Node order when construction `Line`s. This resulted in some exception throwing in OSMesa.

### DIFF
--- a/src/main/scala/vectorpipe/osm/Element.scala
+++ b/src/main/scala/vectorpipe/osm/Element.scala
@@ -99,7 +99,7 @@ case class Way(
   /** Is it a Polyline, but not an "Area" even if closed? */
   def isLine: Boolean = !isClosed || (!isArea && isHighwayOrBarrier)
 
-  def isClosed: Boolean = if (nodes.isEmpty) false else nodes(0) === nodes.last
+  def isClosed: Boolean = (nodes.headOption, nodes.lastOption).mapN(_ === _).getOrElse(false)
 
   def isArea: Boolean = data.tagMap.get("area").map(_ === "yes").getOrElse(false)
 
--- a/src/main/scala/vectorpipe/osm/Element.scala
+++ b/src/main/scala/vectorpipe/osm/Element.scala
@@ -5,6 +5,7 @@ import java.time._
 import cats.implicits._
 import geotrellis.vector.{ Extent, Point }
 import io.dylemma.spac._
+import monocle.macros.Lenses
 
 // --- //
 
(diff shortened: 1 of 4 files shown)


src/main/scala/vectorpipe/osm/Element.scala
```diff
@@ -123,10 +124,11 @@ case class Member(
   ref: Long,
   role: String)
 
-case class ElementData(meta: ElementMeta, tagMap: Map[String, String])
+/** All Elements have common attributes, and may also have additional "tags". */
+@Lenses case class ElementData(meta: ElementMeta, tagMap: Map[String, String])
```
fosskers: Auto-generated lenses are placed in the type's companion object.

src/main/scala/vectorpipe/osm/internal/ElementToFeature.scala
```diff
@@ -283,7 +283,7 @@ private[vectorpipe] object ElementToFeature {
     case Vector()  => Vector.empty.pure[LineState]
     case Vector(h) => State.modify[List[OSMLine]](h :: _).map(_ => Vector.empty)
     case v => fuseOne(v) match {
-      case None => State.modify[List[OSMLine]](v.head :: _) >> fuseLines(v.tail)
+      case None => State.modify[List[OSMLine]](v.head :: _) *> fuseLines(v.tail)
```
fosskers: Cats recently deprecated Monadic `>>` in favour of Applicative `*>`.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -102,6 +93,26 @@ private[vectorpipe] object PlanetHistory {
...
+    f: Vector[Point] => G,
+    w: Way,
+    slices: List[(Instant, Map[Long, Node])]
+  ): List[Feature[G, ElementData]] = {
+    slices.foldLeft(Nil: List[Feature[G, ElementData]]) { case (acc, (i, ns)) =>
+      /* If a Node were deleted that the Way still expected, then no Line
+       * should be formed.
+       */
+      val points: Option[Vector[Point]] =
+        w.nodes.map(id => ns.get(id).map(n => Point(n.lon, n.lat))).sequence
```
fosskers: This line in particular fixes the bug.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -52,6 +53,9 @@ private[vectorpipe] object PlanetHistory {
       ps: List[OSMPolygon]
     ): (List[OSMLine], List[OSMPolygon]) = ws match {
       case Nil => (ls, ps)
+      /* OSM Extracts can produce degenerate Ways. These two checks guard against those. */
+      case w :: _ if w.nodes.length < 4 && w.isClosed => (ls, ps)
+      case w :: _ if w.nodes.length < 2 => (ls, ps)
```
fosskers: Guard against degenerate Ways.

src/main/scala/vectorpipe/osm/Element.scala
```diff
@@ -123,10 +124,11 @@ case class Member(
...
 
 /** All Element types have these attributes in common. */
-case class ElementMeta(
+@Lenses case class ElementMeta(
```
moradology: I prefer this on a separate line, but I don't think this preference rises to the level of a standard

fosskers: The Monocle author does it that way in his docs: https://julien-truffaut.github.io/Monocle/optics/lens.html

PR #37, opened 2017-11-08
Minor API Tweaks

fosskers:
From suggestions from @mojodna .

The main change here is the combining of `ElementMeta` and `ElementData`. There wasn't a reason why they had to be apart, and the API is smaller now.

### DIFF
--- a/src/main/scala/vectorpipe/osm/Element.scala
+++ b/src/main/scala/vectorpipe/osm/Element.scala
@@ -100,11 +100,8 @@ private[vectorpipe] object Element {
 
   def isArea: Boolean = meta.tags.get("area").map(_ === "yes").getOrElse(false)
 
-  def isHighwayOrBarrier: Boolean = {
-    val tags: Set[String] = meta.tags.keySet
-
-    tags.contains("highway") || tags.contains("barrier")
-  }
+  private[this] def isHighwayOrBarrier: Boolean =
+    meta.tags.contains("highway") || meta.tags.contains("barrier")
 }
 
 @Lenses case class Relation(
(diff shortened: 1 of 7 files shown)


src/main/scala/vectorpipe/osm/Element.scala
```diff
@@ -124,15 +120,14 @@ case class Member(
   ref: Long,
   role: String)
 
-/** All Elements have common attributes, and may also have additional "tags". */
-@Lenses case class ElementData(meta: ElementMeta, tagMap: Map[String, String])
```
fosskers: `ElementMeta` replaces `ElementData` in all cases. There wasn't really a reason to have them separate in the first place.

src/main/scala/vectorpipe/osm/Element.scala
```diff
@@ -17,7 +17,7 @@ sealed trait Element {
 private[vectorpipe] object Element {
   implicit val tag: Parser[(String, String)] = (
     Parser.forMandatoryAttribute("k") ~
-      Parser.forMandatoryAttribute("v")
+    Parser.forMandatoryAttribute("v")
```
mojodna: `v` may be `""` (and `k`, too, under rare circumstances).

fosskers: What would the `<tag ...>` look like in that case?

mojodna:
```xml
<tag k="" v="" />
```

mojodna: So, depends on what "mandatory" means...

fosskers: Oh, in that case we're all good. `mandatory` here just means that the `k` and `v` *attributes* must exist or the parse should fail.

PR #38, opened 2017-11-14
Yield all (simple) Historical Features

fosskers:
### TODO

- [x] Return constructed Polygons
- [x] Return standalone Nodes as GeoTrellis Points
- [x] Fixed bug which caused most historical nodes to be dropped
- [x] Microsite still builds

**Note:** This doesn't yet include complex Polygons formed from relations - only simple ones from Ways. These are the vast majority.

### DIFF
--- a/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
+++ b/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
@@ -32,14 +32,23 @@ private[vectorpipe] object PlanetHistory {
     ways: RDD[(Long, Way)]
   ): RDD[(Long, (Iterable[Way], Iterable[Node]))] = {
 
+    /* Forgive the `.distinct` here. If we don't do that, there will be Way ID duplication
+     * in the first cogroup below.
+     */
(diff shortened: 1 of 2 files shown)


src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -13,30 +13,64 @@ import vectorpipe.osm._
...
     val nodesToWayIds: RDD[(Node, Iterable[Long])] =
       nodes
-        .cogroup(ways.flatMap { case (wayId, way) => way.nodes.map { nodeId => (nodeId, wayId) } })
+        .cogroup(nodeIdsToWayIds)
         .flatMap {
-          /* ASSUMPTION: `nodes` contains distinct elements */
-          case (_, (nodes, wayIds)) => nodes.headOption.map(n => (n, wayIds))
+          /* The `.distinct` above ensures that `wayIds` here will contain unique values. */
+          case (_, (ns, wayIds)) => ns.map(n => (n, wayIds))
         }
```
fosskers: This function is called twice. Should we call `.cache` on it?

PR #40, opened 2017-11-20
Misc typos

mojodna:


src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -148,7 +148,7 @@ private[vectorpipe] object PlanetHistory {
     slices: List[(ElementMeta, Map[Long, Node])]
   ): List[Feature[G, ElementMeta]] = {
     slices.foldLeft(Nil: List[Feature[G, ElementMeta]]) { case (acc, (m, ns)) =>
-      /* If a Node were deleted that the Way still expected, then no Line
```
fosskers: `were` is proper use of the subjunctive mood here.

src/main/tut/usage/osm.md
```diff
@@ -65,5 +65,5 @@ osm.fromORC(path) match {
 ss.stop()
 ```
 
-This approach will be particularly efficient when ran on an EMR cluster, since
+This approach will be particularly efficient when run on an EMR cluster, since
```
fosskers: Well shucks, turns out `run` is the correct past participle! [I never knew.](https://www.usingenglish.com/reference/irregular-verbs/run.html)

> I have run many marathons.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
mojodna: You'd think that I'd know that after le subjonctif was the thing I studied most in French class...

PR #41, opened 2017-11-20
Credit subsequent edits

mojodna:
Nodes participating in the way being reconstructed were previously being credited rather than using metadata on the way itself. This resulted in way edits being credited to incorrect changesets (often much older).

This also reduces the set of candidate nodes associated with the way being reconstructed.

### DIFF
--- a/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
+++ b/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
@@ -94,16 +94,13 @@ private[vectorpipe] object PlanetHistory {
         /* Y2K bug here, except it won't manifest until the end of the year 1 billion AD */
         val nextTime: Instant = rest.headOption.map(_.meta.timestamp).getOrElse(Instant.MAX)
 
-        /* Nodes associated with this way */
-        val ns = nodes.filter(n => w.nodes.contains(n.meta.id))
-
         /* Each full set of Nodes that would have existed for each time slice. */
         val allSlices: List[(ElementMeta, Map[Long, Node])] =
-          changedNodes(w.meta.timestamp, nextTime, ns)
+          changedNodes(w.meta.timestamp, nextTime, w.nodes)
             .groupBy(_.meta.timestamp)
             .toList
             .sortBy { case (i, _) => i }
-            .scanLeft((w.meta, recentNodes(w, ns))) { case ((prevMeta, p), (_, changes)) =>
+            .scanLeft((w.meta, recentNodes(w, w.nodes))) { case ((prevMeta, p), (_, changes)) =>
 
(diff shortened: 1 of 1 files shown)


src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -94,13 +94,16 @@ private[vectorpipe] object PlanetHistory {
         /* Y2K bug here, except it won't manifest until the end of the year 1 billion AD */
         val nextTime: Instant = rest.headOption.map(_.meta.timestamp).getOrElse(Instant.MAX)
 
+        /* Nodes associated with this way */
+        val ns = nodes.filter(n => w.nodes.contains(n.meta.id))
```
fosskers: `ns  == nodes` here. Apologies for not making this clearer: the `ways` and `nodes` lists coming into `linesAndPolys` are already associated. The Ways are versions of *one* Way, and the Nodes are all versions of all Nodes ever associated with that Way. 

Given that, is the PR still necessary?

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -111,14 +114,19 @@ private[vectorpipe] object PlanetHistory {
 
               /* The user who changed the Nodes in this timeslice is credited
                * with the creation of the `Line`, even if they weren't the one
-               * who created the Way to begin with.
+               * who created the Way to begin with (unless the Node previously
+               * existed, in which case the Way's author should be credited).
```
fosskers: Ok, point taken.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
mojodna: They're already associated, but not with individual ways (as I read it); does reducing the set of candidate nodes when reconstructing this particular way help?

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
mojodna: Without this, I was seeing geometries being incorrectly credited.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
mojodna: Forgive my ignorance about Spark, but does the `filter` not provide a benefit by reducing the number of times the full `nodes` list needs to be iterated over in `recentNodes()`?

fosskers: We're not in Spark-land here. Notice the inputs are just `List`.

mojodna:
`filter` removed now that I understand some additional details.

PR #42, opened 2017-11-20
Minor versions

mojodna:
This assigns a `minorVersion` to each feature.

This is useful for a) distinguishing between edits of the element itself and those referred to and b) providing an additional component to use as a key when referring to specific geometry versions.

(This incorporates #41)


fosskers:
Thanks!

PR #43, opened 2017-12-04
Separated Feature RDDs

fosskers:
### TODO

- [x] `Features` type
- [x] Make both snapshot and historical APIs use `Features`
- [x] Fix microsite

### Motivation

This PR adds the `Features` type, a simple wrapper around the RDDs of various `Geometry` subtypes produced by the Elements-to-Features process. It's often the case that we want to inspect only a subset of the recreated features ("roads only", for instance). In these cases, we don't want to bother with Spark processing all the extra Points and Polygons. With the `Features` wrapper, it no longer will.

`Features` exposes the function `geometries: Features => RDD[OSMFeature]` as the glue to the rest of the VectorTile creation process, since the next step wants generic `OSMFeature`s.

The results of this PR should improve performance for OSMesa analytics generation, as well as other Azavea ingests that involve OSM data.

### DIFF
--- a/src/main/scala/vectorpipe/osm/Features.scala
+++ b/src/main/scala/vectorpipe/osm/Features.scala
@@ -16,7 +16,7 @@ case class Features(
   multiPolys: RDD[Feature[MultiPolygon, ElementMeta]]
 ) {
   /** Flatten the separated features into a single `RDD` of their supertype, `Geometry`. */
-  def geometries: RDD[Feature[Geometry, ElementMeta]] = points.sparkContext.union(
+  def geometries: RDD[OSMFeature] = points.sparkContext.union(
     points.map(identity), lines.map(identity), polygons.map(identity), multiPolys.map(identity)
   )
 }
--- a/src/main/scala/vectorpipe/osm/Features.scala
+++ b/src/main/scala/vectorpipe/osm/Features.scala
@@ -1,0 +1,22 @@
+package vectorpipe.osm
+
+import geotrellis.vector._
+import org.apache.spark.rdd.RDD
+
+// --- //
+
+/** A meaningful wrapping around the various geometry types that can be produced
+  * from the OSM-to-Features conversion. By keeping each Geometry subtype separate,
+  * the user can discard types they don't need.
+  */
+case class Features(
+  points:     RDD[Feature[Point, ElementMeta]],
+  lines:      RDD[Feature[Line, ElementMeta]],
+  polygons:   RDD[Feature[Polygon, ElementMeta]],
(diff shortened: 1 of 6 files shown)


src/main/scala/vectorpipe/osm/Features.scala
```diff
@@ -0,0 +1,22 @@
...
+  */
+case class Features(
+  points:     RDD[Feature[Point, ElementMeta]],
+  lines:      RDD[Feature[Line, ElementMeta]],
+  polygons:   RDD[Feature[Polygon, ElementMeta]],
+  multiPolys: RDD[Feature[MultiPolygon, ElementMeta]]
+) {
+  /** Flatten the separated features into a single `RDD` of their supertype, `Geometry`. */
+  def geometries: RDD[OSMFeature] = points.sparkContext.union(
+    points.map(identity), lines.map(identity), polygons.map(identity), multiPolys.map(identity)
```
mojodna: What's the purpose of `.map(identity)`?

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -30,7 +27,17 @@ private[vectorpipe] object PlanetHistory {
     val lines:  RDD[OSMLine]    = lap.keys.flatMap(identity)
     val polys:  RDD[OSMPolygon] = lap.values.flatMap(identity)
 
-    (points, lines, polys)
+    /* Depending on the dataset used, `Way` data may be incomplete. That is,
+     * the local version of a Way may have fewer Node references that the original
```
mojodna: **than** the original

src/main/scala/vectorpipe/osm/Features.scala
fosskers: Spark RDDs are "invariant", you can't willy-nilly put things of different datatypes (or mixed super/subtypes) in them. `.map(identity)` is a trick to cast each of the subtypes up to `Geometry` to get around this.

mojodna:
Can you elaborate on this?

> With the Features wrapper, it no longer will.

I _think_ this means that if I provide a filter that limits ways to those containing a particular tag (`highway=*`) it will return an RDD that will skip processing of non-highway Ways when I apply additional functions to it. Is this accurate? (If so, what does the Spark code that would do this look like?)

fosskers:
What that lines means is that if you only cared about the roads, you'd:
```scala
val roadsOnly: RDD[(Long, Way)] = ways.filter { case (_, w) => w.meta.tags.contains("highway") }

val fs: Features = osm.snapshotFeatures(..., roadsOnly, ...)

val roads: RDD[Feature[Line, ElementMeta]] = fs.lines
```

PR #44, opened 2017-12-05
Remove `hadoop-aws` dep

fosskers:
And `useS3` along with it. It wasn't providing an advantage, and muddles the dependency graph.


moradology:
+1

PR #46, opened 2017-12-06
OSM Performance Improvements

fosskers:
### TODO

- [x] Caching with ORC handling
- [x] Caching in Planet History
- [x] Caching in Planet snapshot

### DIFF
--- a/src/main/scala/vectorpipe/osm/package.scala
+++ b/src/main/scala/vectorpipe/osm/package.scala
@@ -9,6 +9,7 @@ import geotrellis.vector._
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql._
+import org.apache.spark.storage.StorageLevel
 import vectorpipe.osm.internal.{ ElementToFeature => E2F, PlanetHistory }
 
 // --- //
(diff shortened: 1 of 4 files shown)


src/main/scala/vectorpipe/osm/package.scala
```diff
@@ -43,6 +44,8 @@ package object osm {
     * read out RDDs of each [[Element]] type.
     */
   def fromDataFrame(data: DataFrame): (RDD[(Long, Node)], RDD[(Long, Way)], RDD[(Long, Relation)]) = {
+    data.persist(StorageLevel.MEMORY_AND_DISK_SER)
```
mojodna: Depending on what the `DataFrame` is backed by, will this prevent predicate push-down (supported by ORC, Parquet, and others) and cache the *entire* dataset, whether or not it needs to be fully read?

fosskers: I don't know. All I'm scared of is the data being downloaded three times for when each of the Nodes, Ways, and Relations are being read.

mojodna: Watching the Spark UI during this process leads me to believe that without this it will perform **3** full table scans, but will be able to skip stripes via predicate push-down.

_goes to do some reading..._

fosskers: I'm going to stare at a Spark UI myself now to try and divine the truth.

mojodna: A cursory look around the internet doesn't provide any clues about what makes sense here.

fosskers: With `historicalFeatures` running on a snapshot instead, and the current caching behaviour here, I was able to read, convert, and rasterize the current worldly snapshot in 2.5 hours.

mojodna: Sweet! How large a cluster?

PR #53, opened 2018-01-09
Set the StorageLevel Used When Deriving Features

### LINKED ISSUE #52: Allow the User to Specify the StorageLevel of Internal, Intermediate RDDs
When getting the `Features` from an OSM source, there are internal `RDD`s that hold intermediate data that are cached (see [here](https://github.com/geotrellis/vectorpipe/blob/master/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala#L54) and [here](https://github.com/geotrellis/vectorpipe/blob/master/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala#L64-L69)). The `StorageLevel` used for both of these `RDD`s is `MEMORY_ONLY`. However, this causes problems when large amounts of data need to be cached. When this occurs, any spill over is written to disk which slows down the whole process. It would be nice to be able to specify a `StorageLevel` or none at all when reading in `Features`.

jbouffard:
This PR adds the optional parameter, `storageLevel` to the `features` method so that the user can set the `StorageLevel` in the internal. intermediate `RDD`s.

This PR resolves #52

### DIFF
--- a/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
+++ b/src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
@@ -15,9 +15,13 @@ import vectorpipe.osm._
 private[vectorpipe] object PlanetHistory {
 
   /** For all given Ways, associate with their Nodes to produce GeoTrellis Lines and Polygons. */
-  def features(nodes: RDD[(Long, Node)], ways: RDD[(Long, Way)]): Features = {
+  def features(
+    nodes: RDD[(Long, Node)],
+    ways: RDD[(Long, Way)],
+    storageLevel: StorageLevel
+  ): Features = {
 
-    val (loneNodes, edges) = joinedWays(nodes, ways)
+    val (loneNodes, edges) = joinedWays(nodes, ways, storageLevel)
 
     val lap: RDD[(List[OSMLine], List[OSMPolygon])] = edges.map { case (_, (ws, ns)) =>
       linesAndPolys(ws.toList, ns.toList)
@@ -47,11 +51,12 @@ private[vectorpipe] object PlanetHistory {
     */
   private[this] def joinedWays(
     nodes: RDD[(Long, Node)],
-    ways: RDD[(Long, Way)]
+    ways: RDD[(Long, Way)],
+    storageLevel: StorageLevel
   ): (RDD[(Long, Node)], RDD[(Long, (Iterable[Way], Iterable[Node]))]) = {
 
-    /* This is called twice below, so we need to cache is, or else we'd repeat work */
-    val cachedWays: RDD[(Long, Way)] = ways.persist(StorageLevel.MEMORY_ONLY)
+    /* This is called twice below, so in most cases we should cache it, or else we'd repeat work */
+    val cachedWays: RDD[(Long, Way)] = ways.persist(storageLevel)
 
     /* Forgive the `.distinct` here. If we don't do that, there will be Way ID duplication
      * in the first cogroup below.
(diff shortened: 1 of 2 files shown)


src/main/scala/vectorpipe/osm/package.scala
```diff
@@ -184,7 +185,12 @@ package object osm {
     *
     * @group conversion
     */
-  def features(ns: RDD[(Long, Node)], ws: RDD[(Long, Way)], rs: RDD[(Long, Relation)]): Features = {
-    PlanetHistory.features(ns, ws)
+  def features(
+    ns: RDD[(Long, Node)],
+    ws: RDD[(Long, Way)],
+    rs: RDD[(Long, Relation)],
+    storageLevel: Option[StorageLevel] = Some(StorageLevel.MEMORY_ONLY)
```
fosskers: Since it's a default arg already, it would probably result in less code overall if it read:
```scala
storageLevel: StorageLevel = StorageLevel.MEMORY_ONLY
```
Then you wouldn't have to deal with threading the `Option` through.

Unless there's a case where someone would specifically want to do *no* persisting? I can't see that being the typical case.

jbouffard: Yeah, the reason I made it an `Option` was in case someone did want any persisting. I'm not sure how often that situation would arise, but I thought it would be good to have the option available.

lossyrob: There are situations where persistence is not desired...if I have a very large dataset on a memory constrained cluster, certain caching strategies could cause the job to always fail with out of memory errors. In that case I would want to turn off persistence to try and make the job succeed at the sacrifice of having the job run slower.

echeipesh: There is `StorageLevel.NONE` that can be used as default: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/storage/StorageLevel.scala#L153 that would get rid of the option.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
```diff
@@ -61,12 +70,18 @@ private[vectorpipe] object PlanetHistory {
         .distinct
```
echeipesh: This is a little out of context but why is `.distinct` required here? Is it that a way can have multiple references to a single node or that there can somehow be two way records with same ID that may potentially reference the same node ?

src/main/scala/vectorpipe/osm/package.scala
fosskers: Some history of why we're handling this at all: during the Tobler work, @echeipesh and I had to run the ingest umpteen times with very specific memory settings, partition settings, and persist/unpersist calls in many places to get the job to work. At that time, setting these persists in VectorPipe to `MEMORY_ONLY` was the only thing that ended up working, since `MEMORY_ONLY` persisted blocks are expected to be ejected automatically if later stages need the memory.

src/main/scala/vectorpipe/osm/internal/PlanetHistory.scala
fosskers: > ...be two way records with same ID that may potentially reference the same node

Precisely, the `.distinct` has to remain or the function will return an incorrect result.

jbouffard:
@fosskers @echeipesh I updated this PR so that now `storageLevel: StorageLevel` instead of `storageLevel: Option[StorageLevel`. Is there anything else that needs to changed/fixed?

src/main/scala/vectorpipe/osm/package.scala
```diff
@@ -184,7 +185,12 @@ package object osm {
     *
     * @group conversion
     */
-  def features(ns: RDD[(Long, Node)], ws: RDD[(Long, Way)], rs: RDD[(Long, Relation)]): Features = {
-    PlanetHistory.features(ns, ws)
+  def features(
```
fosskers: A note in the docstrings here about the various Storage Levels might be handy, unless that's something that's better addressed in the Microsite docs.

PR #54, opened 2018-05-29
osmesa Code Migration

jbouffard:
This PR seeks to migrate the code dealing with creating JTS DataFrames from osmesa to vectorpipe. By doing this, DataFrame creating will be more performant and easier. In addition to the code migration, new API will developed so that the new logic can be accessed in an easy way.


mojodna:
@jbouffard have a look at https://github.com/azavea/osmesa/pull/63 for relation and way assembly improvements (I'm planning to have another pass to see if I can improve the memory footprint of final assembly, currently done with `mapPartitions`, using `groupByKey` / `mapGroups`). The UDAFs didn't work as well as I'd originally hoped.

jbouffard:
Hey @mojodna! Thanks for letting me know about that PR. I've since updated this one to reflect the updates you made in osmesa. There was a noticeable improvement in performance once I made the changes.

However, I think a bug may have been introduced as I'm getting an error when trying to save geometries in the `ExtractMultiPolygons` script. It occurred both when using osmesa and vectorpipe. I created an issue for it here https://github.com/azavea/osmesa/issues/66

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,636 @@
...
+            case "" if line.isClosed && line.getNumPoints >= 4 =>
+              (c :+ GeomFactory.factory.createPolygon(line.getCoordinateSequence), p)
+            case "" => (c, p :+ line)
+            case _ => (c, p)
+          }
+      }
+
+      try {
+        val rings = complete ++ formRings(partial.sortWith(_.getNumPoints > _.getNumPoints))
+        val preparedRings = rings
```
jbouffard: One of the changes I made from @mojodna's original code was changing `preparedRings` from a `PreparedGeometry` to just a `Geometry`. I'm not familiar with `PreparedGeometries`, so I'm not sure if the change has a significant impact or not. This is the original code: https://github.com/azavea/osmesa/blob/master/src/ingest/src/main/scala/osmesa/functions/osm/package.scala#L509

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,636 @@
...
+        val (dissolvedOuters, addlInners) = dissolveRings(classifiedOuters)
+        val (dissolvedInners, addlOuters) = dissolveRings(classifiedInners.map(_.getExteriorRing.getCoordinateSequence).map(GeomFactory.factory.createPolygon) ++ addlInners)
+
+        val (polygons, _) = (dissolvedOuters ++ addlOuters)
+          // sort by size (descending) to use rings as part of the largest available polygon
+          .sortWith(_.getArea > _.getArea)
+          // only use inners once if they're contained by multiple outer rings
+          .foldLeft((Vector.empty[Polygon], dissolvedInners)) {
+          case ((ps, is), outer) =>
+            val preparedOuter = outer
```
jbouffard: Same as the above comment where `preparedOut` is now just a `Geometry`. Original code: https://github.com/azavea/osmesa/blob/master/src/ingest/src/main/scala/osmesa/functions/osm/package.scala#L535

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,767 @@
...
+    */
+  def addUserMetadata(geoms: DataFrame, changesets: DataFrame): DataFrame = {
+    import geoms.sparkSession.implicits._
+
+    geoms
+      .join(changesets.select('id as 'changeset, 'uid, 'user), Array("changeset"))
+  }
+
+  /*
+  def geometriesByRegion(nodeGeoms: DataFrame, wayGeoms: DataFrame): DataFrame = {
```
jbouffard: I commented this out because I wasn't sure if this was a feature we needed right now and because I have a few questions on it. I can re-implement it though if it's desired for this PR.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
lossyrob: Using prepared geometries is actually an important optimization. A prepared geometry is one in which predicates like 'intersects' runs much faster than the unprepared version.

As a general rule, I'd recommend against taking out or changing functionality that isn't well understood, without consulting with the author or getting a clear understanding of why the choices were made - those choices could come from hours of debugging and optimizing, or they could be mistakes, but it's important to know which. Not knowing is a common way I've seen regressions happen!

Thanks for bringing that point up.

fosskers:
Hey @jbouffard , Codacy is set up for this repo, so by hitting the `Details` link there you can see a lot of suggestions for improving the PR. Generally low-hanging fruit that increase long-term maintainability, so they're free wins pretty much.

build.sbt
```diff
@@ -22,7 +23,25 @@ val common = Seq(
...
     "-Ywarn-unused-import",
     "-Ywarn-dead-code",
-    "-Ywarn-numeric-widen"
+    "-Ywarn-numeric-widen",
+    "-language:implicitConversions",
+    "-language:reflectiveCalls",
+    "-language:higherKinds",
+    "-language:postfixOps",
+    "-language:existentials",
+    "-language:experimental.macros"
```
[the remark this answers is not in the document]
jbouffard: Those are just cargo-culted since I wasn't sure what I'd be needing. I can remove those.

build.sbt
```diff
@@ -22,7 +23,25 @@ val common = Seq(
...
+  ),
+
+  resolvers ++= Seq(
+    Resolver.bintrayRepo("lonelyplanet", "maven"),
+    Resolver.bintrayRepo("kwark", "maven"), // Required for Slick 3.1.1.2, see https://github.com/azavea/raster-foundry/pull/1576
```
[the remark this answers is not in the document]
jbouffard: I don't think so. That's another thing I will remove.

project/Dependencies.scala
```diff
@@ -0,0 +1,15 @@
...
+  val gtS3         = "org.locationtech.geotrellis" %% "geotrellis-s3"         % Version.geotrellis
+  val geomesa      = "org.locationtech.geomesa"    %% "geomesa-spark-jts"     % Version.geomesa
+  val sparkHive    = "org.apache.spark"            %% "spark-hive"            % Version.spark
+  val spire        = "org.spire-math"              %% "spire"                 % Version.spire
+  val cats         = "org.typelevel"               %% "cats-core"             % Version.cats
+  val decline      = "com.monovore"                %% "decline"               % Version.decline
+  val scalaTest    = "org.scalatest"               %% "scalatest"             % Version.scalaTest
+  val monocleCore  = "com.github.julien-truffaut"  %% "monocle-core"          % Version.monocle
+  val monocleMacro = "com.github.julien-truffaut"  %% "monocle-macro"         % Version.monocle
+}
```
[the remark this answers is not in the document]
jbouffard: Not that I know of. I just did it to keep in line with what GT does. I think it would be easier to put them in the `build` so I'll do that as well.

project/assembly.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.5")
```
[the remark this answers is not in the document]
jbouffard: I've been using assemblies to test this PR, but I'm not sure if removing this will impact anything.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
mojodna: I'll take a look at these. I removed the preparedness while switching to use JTS for dissolving rings.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
mojodna: I don't think this is needed now.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
jbouffard: I see. Next time I will look into the type more and/or ask someone for their opinion. I have since reverted my changes and now `PreparedGeometries` are used. Thanks for explaining what those types are.

mojodna:
@jbouffard there are a few additional changes to `ProcessOSM` that should be pulled from https://github.com/azavea/osmesa/pull/65

jbouffard:
@mojodna Thanks for letting me know!

mojodna:
`ExtractMultiPolygons` was fixed by https://github.com/azavea/osmesa/pull/67

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,636 @@
...
+      }
+      val ref = row.getAs[Long]("ref")
+      val role = row.getAs[String]("role")
+
+      Row(t, ref, role)
+    }
+
+  val compressMemberTypes: UserDefinedFunction = udf(_compressMemberTypes, MemberSchema)
+
+  class AssemblyException(msg: String) extends Exception(msg)
```
mojodna: @jbouffard from here down would benefit from some cleanup / refactoring / extraction.

jbouffard: Okay, I'll take a look and see where I can improve things.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
mojodna: Ah, I didn't realize that I left the `PreparedGeometries` in. Great.

mojodna:
@jbouffard before this is merged, can you add a comment with a diff against the corresponding files in OSMesa `master` so we can put some more eyes on the changes?

src/main/scala/vectorpipe/osm/internal/functions/package.scala
lossyrob: @mojodna are prepared geoms not required here then?

mojodna: They are required. I thought I'd removed them on the OSMesa side when switching to JTS after not reading @jbouffard's comment about having changed them here.

lossyrob: Ah, gotcha.

jbouffard:
@mojodna This is the diff for `ProcessOSM.scala`

```diff
diff --git a/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala b/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
index 6e27091..3aa2438 100644
--- a/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
+++ b/src/ingest/src/main/scala/osmesa/ingest/ProcessOSM.scala
@@ -1,11 +1,14 @@
-package osmesa
+package vectorpipe.osm.internal
 
-import java.io._
 import java.sql.Timestamp
 
-import geotrellis.vector._
-import geotrellis.vector.io._
-import geotrellis.vector.io.json._
+import org.locationtech.geomesa.spark.jts._
+import org.apache.spark.sql.jts.GeometryUDT
+
+import com.vividsolutions.jts.geom._
+
+import geotrellis.vector.{Extent, GeomFactory}
+
 import org.apache.log4j.Logger
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.encoders.RowEncoder
@@ -13,10 +16,10 @@ import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
 import org.apache.spark.sql.expressions.Window
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types._
-import osmesa.functions._
-import osmesa.functions.osm._
-import osmesa.ingest.util.Caching
-import spray.json._
+
+import vectorpipe.osm.internal.functions._
+import vectorpipe.osm.internal.util.Caching
+
 
 object ProcessOSM {
   val NodeType: Byte = 1
@@ -31,7 +34,7 @@ object ProcessOSM {
       StructField("id", LongType, nullable = false) ::
       StructField("version", IntegerType, nullable = false) ::
       StructField("updated", TimestampType, nullable = false) ::
-      StructField("geom", BinaryType) ::
+      StructField("geom", GeometryUDT) ::
       Nil)
 
   val BareElementEncoder: Encoder[Row] = RowEncoder(BareElementSchema)
@@ -44,7 +47,7 @@ object ProcessOSM {
       StructField("minorVersion", IntegerType, nullable = false) ::
       StructField("updated", TimestampType, nullable = false) ::
       StructField("validUntil", TimestampType) ::
-      StructField("geom", BinaryType) ::
+      StructField("geom", GeometryUDT) ::
       Nil)
 
   val TaggedVersionedElementEncoder: Encoder[Row] = RowEncoder(TaggedVersionedElementSchema)
@@ -56,7 +59,7 @@ object ProcessOSM {
       StructField("minorVersion", IntegerType, nullable = false) ::
       StructField("updated", TimestampType, nullable = false) ::
       StructField("validUntil", TimestampType) ::
-      StructField("geom", BinaryType) ::
+      StructField("geom", GeometryUDT) ::
       Nil)
 
   val VersionedElementEncoder: Encoder[Row] = RowEncoder(VersionedElementSchema)
@@ -248,7 +251,7 @@ object ProcessOSM {
       .select(
         lit(NodeType) as '_type,
         'id,
-        ST_Point('lon, 'lat) as 'geom,
+        st_makePoint('lon, 'lat) as 'geom,
         'tags,
         'changeset,
         'updated,
@@ -339,31 +342,31 @@ object ProcessOSM {
                   Option(row.get(row.fieldIndex("lat"))).map(_.asInstanceOf[Float]).getOrElse(Float.NaN))
             } match {
               // no coordinates provided
-              case coords if coords.isEmpty => Some("LINESTRING EMPTY".parseWKT)
+              case coords if coords.isEmpty => Some(GeomFactory.factory.createLineString(Array[Coordinate]()))
               // some of the coordinates are empty; this is invalid
               case coords if coords.exists(Option(_).isEmpty) => None
               // some of the coordinates are invalid
               case coords if coords.exists(_.exists(_.isNaN)) => None
               // 1 pair of coordinates provided
               case coords if coords.length == 1 =>
-                Some(Point(coords.head.head, coords.head.last))
+                Some(GeomFactory.factory.createPoint(new Coordinate(coords.head.head, coords.head.last)))
               case coords => {
-                coords.map(xy => (xy.head.toDouble, xy.last.toDouble)) match {
-                  case pairs => Line(pairs)
-                }
-              } match {
-                case ring if isArea && ring.vertexCount >= 4 && ring.isClosed =>
-                  Some(Polygon(ring))
-                case line => Some(line)
+                val coordinates = coords.map(xy => new Coordinate(xy.head.toDouble, xy.last.toDouble)).toArray
+                val line = GeomFactory.factory.createLineString(coordinates)
+
+                if (isArea && line.getNumPoints >= 4 && line.isClosed)
+                  Some(GeomFactory.factory.createPolygon(line.getCoordinateSequence))
+                else
+                  Some(line)
               }
             }
 
-              val wkb = geom match {
-                case Some(g) if g.isValid => g.toWKB(4326)
+              val geometry = geom match {
+                case Some(g) if g.isValid => g
                 case _ => null
               }
 
-            new GenericRowWithSchema(Array(changeset, id, version, updated, wkb), BareElementSchema): Row
+            new GenericRowWithSchema(Array(changeset, id, version, updated, geometry), BareElementSchema): Row
       }
 
     @transient val idAndVersionByUpdated = Window.partitionBy('id, 'version).orderBy('updated)
@@ -463,17 +466,20 @@ object ProcessOSM {
     * @param geoms      DataFrame containing way geometries to use in reconstruction.
     * @return Relations geometries.
     */
-  def reconstructRelationGeometries(_relations: DataFrame, geoms: DataFrame)(implicit cache: Caching = Caching.none,
-                                                                                         cachePartitions: Option[Int] = None): DataFrame = {
+  def reconstructRelationGeometries(
+    _relations: DataFrame,
+    geoms: DataFrame
+  )(implicit cache: Caching = Caching.none, cachePartitions: Option[Int] = None): DataFrame = {
     val relations = preprocessRelations(_relations)
 
     reconstructMultiPolygonRelationGeometries(relations, geoms)
       .union(reconstructRouteRelationGeometries(relations, geoms))
   }
 
-  def reconstructMultiPolygonRelationGeometries(_relations: DataFrame, geoms: DataFrame)(implicit cache: Caching = Caching.none,
-                                                                             cachePartitions: Option[Int] = None)
-  : DataFrame = {
+  def reconstructMultiPolygonRelationGeometries(
+    _relations: DataFrame,
+    geoms: DataFrame
+  )(implicit cache: Caching = Caching.none, cachePartitions: Option[Int] = None): DataFrame = {
     implicit val ss: SparkSession = _relations.sparkSession
     import ss.implicits._
 
@@ -499,6 +505,7 @@ object ProcessOSM {
 
     implicit val encoder: Encoder[Row] = VersionedElementEncoder
 
+    // leverage partitioning (avoids repeated (de-)serialization of merged coordinate arrays)
     val relationGeoms = members
         .groupByKey { row =>
           (row.getAs[Long]("changeset"), row.getAs[Long]("id"), row.getAs[Integer]("version"), row.getAs[Integer]
@@ -509,11 +516,11 @@ object ProcessOSM {
             val members = rows.toVector
             val types = members.map(_.getAs[Byte]("type"))
             val roles = members.map(_.getAs[String]("role"))
-            val geoms = members.map(_.getAs[Array[Byte]]("geom"))
+            val geoms = members.map(_.getAs[Geometry]("geom"))
 
-            val wkb = buildMultiPolygon(id, version, updated, types, roles, geoms).orNull
+            val geom = buildMultiPolygon(id, version, updated, types, roles, geoms).orNull
 
-            new GenericRowWithSchema(Array(changeset, id, version, minorVersion, updated, validUntil, wkb),
+            new GenericRowWithSchema(Array(changeset, id, version, minorVersion, updated, validUntil, geom),
               VersionedElementSchema): Row
         }
 
@@ -533,9 +540,10 @@ object ProcessOSM {
         'minorVersion)
   }
 
-  def reconstructRouteRelationGeometries(_relations: DataFrame, geoms: DataFrame)(implicit cache: Caching = Caching.none,
-                                                                             cachePartitions: Option[Int] = None)
-  : DataFrame = {
+  def reconstructRouteRelationGeometries(
+    _relations: DataFrame,
+    geoms: DataFrame
+  )(implicit cache: Caching = Caching.none, cachePartitions: Option[Int] = None): DataFrame = {
     implicit val ss: SparkSession = _relations.sparkSession
     import ss.implicits._
 
@@ -571,18 +579,18 @@ object ProcessOSM {
           val members = rows.toVector
           val types = members.map(_.getAs[Byte]("type"))
           val roles = members.map(_.getAs[String]("role"))
-          val geoms = members.map(_.getAs[Array[Byte]]("geom"))
+          val geoms = members.map(_.getAs[Geometry]("geom"))
 
           buildRoute(id, version, updated, types, roles, geoms) match {
             case Some(components) =>
               components.map {
-                case ("", wkb) =>
+                case ("", geom) =>
                   // no role
                   new GenericRowWithSchema(Array(changeset, id, Map(), version, minorVersion, updated,
-                    validUntil, wkb), TaggedVersionedElementSchema): Row
-                case (role, wkb) =>
+                    validUntil, geom), TaggedVersionedElementSchema): Row
+                case (role, geom) =>
                   new GenericRowWithSchema(Array(changeset, id, Map("role" -> role), version, minorVersion,
-                    updated, validUntil, wkb), TaggedVersionedElementSchema): Row
+                    updated, validUntil, geom), TaggedVersionedElementSchema): Row
               }
             case None =>
               // no geometry
@@ -621,6 +629,7 @@ object ProcessOSM {
       .join(changesets.select('id as 'changeset, 'uid, 'user), Array("changeset"))
   }
 
+  /*
   def geometriesByRegion(nodeGeoms: DataFrame, wayGeoms: DataFrame): DataFrame = {
     import nodeGeoms.sparkSession.implicits._
 
@@ -742,7 +751,6 @@ object ProcessOSM {
       .where('country.isNotNull)
       .groupBy('changeset)
       .agg(collect_set('country) as 'countries)
-
   }
-
+  */
 }
```

jbouffard:
And this is the diff for `package.scala`:


```diff
diff --git a/src/ingest/src/main/scala/osmesa/functions/osm/package.scala b/src/ingest/src/main/scala/osmesa/functions/osm/package.scala
index e7c18b9..79b166a 100644
--- a/src/ingest/src/main/scala/osmesa/functions/osm/package.scala
+++ b/src/ingest/src/main/scala/osmesa/functions/osm/package.scala
@@ -1,23 +1,49 @@
-package osmesa.functions
+package vectorpipe.osm.internal
+
+import java.math.BigDecimal
+
+import vectorpipe.osm.internal.ProcessOSM._
+
+import geotrellis.vector.{GeomFactory, Polygon => GTPolygon}
+
+import org.apache.spark.sql.expressions.UserDefinedFunction
+import org.apache.spark.sql.functions.udf
 
 import java.sql.Timestamp
 
-import com.google.common.collect.{Range, RangeMap, TreeRangeMap}
 import com.vividsolutions.jts.geom
 import com.vividsolutions.jts.geom._
-import geotrellis.vector.io._
-import geotrellis.vector.{Line, MultiLine, MultiPolygon, Polygon}
+import com.vividsolutions.jts.geom.prep.PreparedGeometryFactory
+
+import com.google.common.collect.{Range, RangeMap, TreeRangeMap}
+
 import org.apache.log4j.Logger
 import org.apache.spark.sql.Row
 import org.apache.spark.sql.expressions.UserDefinedFunction
 import org.apache.spark.sql.functions.udf
 import org.apache.spark.sql.types._
-import osmesa.ProcessOSM._
 
 import scala.annotation.tailrec
 import scala.collection.GenTraversable
 
-package object osm {
+
+package object functions {
+  // Convert BigDecimals to doubles
+  // Reduces size taken for representation at the expense of some precision loss.
+  val asDouble: UserDefinedFunction = udf {
+    Option(_: BigDecimal).map(_.doubleValue).getOrElse(Double.NaN)
+  }
+
+  // Convert BigDecimals to floats
+  // Reduces size taken for representation at the expense of more precision loss.
+  val asFloat: UserDefinedFunction = udf {
+    Option(_: BigDecimal).map(_.floatValue).getOrElse(Float.NaN)
+  }
+
+  val ST_AsText: UserDefinedFunction = udf {
+    Option(_: Geometry).map((_.toText())).getOrElse("")
+  }
+
   // Using tag listings from [id-area-keys](https://github.com/osmlab/id-area-keys) @ v2.8.0.
   private val AreaKeys: Map[String, Map[String, Boolean]] = Map(
     "addr:*" -> Map(),
@@ -132,6 +158,8 @@ package object osm {
 
   private lazy val logger = Logger.getLogger(getClass)
 
+  private val factory = GeomFactory.factory
+
   private val _isArea = (tags: Map[String, String]) =>
     tags match {
       case _ if tags.contains("area") && BooleanValues.contains(tags("area").toLowerCase) =>
@@ -399,7 +427,10 @@ package object osm {
 
   // join segments together into rings
   @tailrec
-  private def formRings(segments: GenTraversable[VirtualCoordinateSequence], rings: Seq[CoordinateSequence] = Vector.empty[CoordinateSequence]): GenTraversable[CoordinateSequence] = {
+  private def formRings(
+    segments: GenTraversable[VirtualCoordinateSequence],
+    rings: Seq[CoordinateSequence] = Vector.empty[CoordinateSequence]
+  ): GenTraversable[CoordinateSequence] = {
     segments match {
       case Nil =>
         rings
@@ -415,24 +446,29 @@ package object osm {
           case None =>
             t.find(line => x == line.getX(line.size - 1) && y == line.getY(line.size - 1)) match {
               case Some(next) =>
-                h.append(new PartialCoordinateSequence(new ReversedCoordinateSequence(next), 1)) +: t.filterNot(line => isEqual(line, next))
+                h.append(
+                  new PartialCoordinateSequence(
+                    new ReversedCoordinateSequence(next), 1
+                  )
+                ) +: t.filterNot(line => isEqual(line, next))
               case None => throw new AssemblyException("Unable to connect segments.")
             }
         }, rings)
     }
   }
 
-  // since GeoTrellis's GeometryFactory is unavailable
   implicit val geometryFactory: GeometryFactory = new geom.GeometryFactory()
 
-  private def formRings(segments: GenTraversable[Line])(implicit geometryFactory: GeometryFactory): GenTraversable[Polygon] =
-    formRings(segments.map(_.jtsGeom.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
+  private def formRings(segments: GenTraversable[LineString])(implicit geometryFactory: GeometryFactory): GenTraversable[Polygon] =
+    formRings(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
       .map(geometryFactory.createPolygon)
-      .map(Polygon(_))
 
   // join segments together
   @tailrec
-  private def connectSegments(segments: GenTraversable[VirtualCoordinateSequence], lines: Seq[CoordinateSequence] = Vector.empty[CoordinateSequence]): GenTraversable[CoordinateSequence] = {
+  private def connectSegments(
+    segments: GenTraversable[VirtualCoordinateSequence],
+    lines: Seq[CoordinateSequence] = Vector.empty[CoordinateSequence]
+  ): GenTraversable[CoordinateSequence] = {
     segments match {
       case Nil =>
         lines
@@ -442,78 +478,99 @@ package object osm {
 
         t.find(line => x == line.getX(0) && y == line.getY(0)) match {
           case Some(next) =>
-            connectSegments(h.append(new PartialCoordinateSequence(next, 1)) +: t.filterNot(line => isEqual(line, next)), lines)
+            connectSegments(
+              h.append(new PartialCoordinateSequence(next, 1)) +: t.filterNot(line => isEqual(line, next)),
+              lines
+            )
           case None =>
             t.find(line => x == line.getX(line.size - 1) && y == line.getY(line.size - 1)) match {
               case Some(next) =>
-                connectSegments(h.append(new PartialCoordinateSequence(new ReversedCoordinateSequence(next), 1)) +: t.filterNot(line => isEqual(line, next)), lines)
+                connectSegments(
+                  h.append(
+                    new PartialCoordinateSequence(new ReversedCoordinateSequence(next), 1)
+                  ) +: t.filterNot(line => isEqual(line, next)),
+                  lines
+                )
               case None => connectSegments(t, lines :+ h)
             }
         }
     }
   }
 
-  private def connectSegments(segments: GenTraversable[Line])(implicit geometryFactory: GeometryFactory): GenTraversable[Line] =
-    connectSegments(segments.map(_.jtsGeom.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
+  private def connectSegments(segments: GenTraversable[LineString])(implicit geometryFactory: GeometryFactory): GenTraversable[LineString] =
+    connectSegments(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
       .map(geometryFactory.createLineString)
-      .map(Line(_))
+
+  private def getHoles(poly: Polygon): List[CoordinateSequence] = {
+    for (i <- 0 until poly.getNumInteriorRing) yield
+      poly.getInteriorRingN(i).getCoordinateSequence
+  }.toList
 
   private def dissolveRings(rings: GenTraversable[Polygon]): (Seq[Polygon], Seq[Polygon]) = {
-    MultiPolygon(rings.toArray).union.asMultiPolygon match {
-      case Some(mp) =>
-        (mp.polygons.map(_.exterior).map(Polygon(_)).toVector, mp.polygons.flatMap(_.holes).map(Polygon(_)).toVector)
-      case None =>
-        (Vector.empty[Polygon], Vector.empty[Polygon])
-    }
+    if (!rings.isEmpty)
+      (
+        rings.map(_.getExteriorRing).map(x => factory.createPolygon(x.getCoordinateSequence)).toVector,
+        rings.flatMap(getHoles).map(factory.createPolygon).toVector
+      )
+    else
+      (Vector.empty[Polygon], Vector.empty[Polygon])
   }
 
   // TODO this (and accompanying functions) doesn't belong here
-  def buildMultiPolygon(id: Long, version: Int, timestamp: Timestamp, types: Seq[Byte], roles: Seq[String], wkbs: Seq[Array[Byte]]): Option[Array[Byte]] = {
-    if (types.zip(wkbs).exists { case (t, g) => t == WayType && Option(g).isEmpty }) {
+  def buildMultiPolygon(
+    id: Long,
+    version: Int,
+    timestamp: Timestamp,
+    types: Seq[Byte],
+    roles: Seq[String],
+    geometries: Seq[Geometry]
+  ): Option[Geometry] = {
+    if (types.zip(geometries).exists { case (t, g) => t == WayType && Option(g).isEmpty }) {
       // bail early if null values are present where they should exist (members w/ type=way)
       logger.debug(s"Incomplete relation: $id @ $version ($timestamp)")
       None
     } else {
-      val bytes = wkbs.map(Option(_)).filter(_.isDefined).map(_.get).map(_.length).sum
 
-      logger.debug(s"$id @ $version ($timestamp) ${bytes.formatted("%,d")} bytes")
-      val geoms = wkbs.map(Option(_).map(_.readWKB) match {
-        case Some(geom: Polygon) => geom.as[Polygon].map(_.exterior)
-        case Some(geom) => geom.as[Line]
-        case None => None
-      })
+      val geoms = geometries.map { geom =>
+        geom match {
+          case p: Polygon => Some(p.getExteriorRing())
+          case l: LineString => Some(l)
+          case _ => None
+        }
+      }
 
-      val vertexCount = geoms.filter(_.isDefined).map(_.get).map(_.vertexCount).sum
+      val vertexCount = geoms.filter(_.isDefined).map(_.get).map(_.getNumPoints).sum
 
-      logger.warn(s"${vertexCount.formatted("%,d")} vertices (${bytes.formatted("%,d")} bytes) from ${types.size} members in $id @ $version ($timestamp)")
+      logger.warn(s"${vertexCount.formatted("%,d")} vertices from ${types.size} members in $id @ $version ($timestamp)")
 
-      val members: Seq[(String, Line)] = roles.zip(geoms)
+      val members: Seq[(String, LineString)] = roles.zip(geoms)
         .filter(_._2.isDefined)
         .map(x => (x._1, x._2.get))
 
-      val (complete, partial) = members.foldLeft((Vector.empty[Polygon], Vector.empty[Line])) {
-        case ((c, p), (role, line: Line)) =>
+      val (complete, partial) = members.foldLeft((Vector.empty[Polygon], Vector.empty[LineString])) {
+        case ((c, p), (role, line: LineString)) =>
           role match {
-            case "outer" if line.isClosed && line.vertexCount >= 4 => (c :+ Polygon(line), p)
-            case "outer" => (c, p :+ line)
-            case "inner" if line.isClosed && line.vertexCount >= 4 => (c :+ Polygon(line), p)
-            case "inner" => (c, p :+ line)
-            case "" if line.isClosed && line.vertexCount >= 4 => (c :+ Polygon(line), p)
-            case "" => (c, p :+ line)
+            case ("outer" | "inner" | "") =>
+              if (line.isClosed && line.getNumPoints >= 4)
+                (c :+ factory.createPolygon(line.getCoordinateSequence), p)
+              else
+                (c, p :+ line)
             case _ => (c, p)
           }
       }
 
       try {
-        val rings = complete ++ formRings(partial.sortWith(_.vertexCount > _.vertexCount))
-        val preparedRings = rings.map(_.prepare)
+        val preparedFactory = new PreparedGeometryFactory
+
+        val rings = complete ++ formRings(partial.sortWith(_.getNumPoints > _.getNumPoints))
+        val preparedRings = rings.map(preparedFactory.create)
 
         // reclassify rings according to their topology (ignoring roles)
-        val (classifiedOuters, classifiedInners) = rings.sortWith(_.area > _.area) match {
+        val (classifiedOuters, classifiedInners) = rings.sortWith(_.getArea > _.getArea) match {
           case Seq(h, t@_ *) => t.foldLeft((Vector(h), Vector.empty[Polygon])) {
             case ((os, is), ring) =>
               // check the number of containing elements
-              preparedRings.count(r => r.geom != ring && r.contains(ring)) % 2 match {
+              preparedRings.count(r => r != ring && r.contains(ring)) % 2 match {
                 // if even, it's an outer ring
                 case 0 => (os :+ ring, is)
                 // if odd, it's an inner ring
@@ -524,21 +581,34 @@ package object osm {
         }
 
         val (dissolvedOuters, addlInners) = dissolveRings(classifiedOuters)
-        val (dissolvedInners, addlOuters) = dissolveRings(classifiedInners.map(_.exterior).map(Polygon(_)) ++ addlInners)
+        val (dissolvedInners, addlOuters) =
+          dissolveRings(
+            classifiedInners
+            .map(_.getExteriorRing.getCoordinateSequence)
+            .map(factory.createPolygon) ++ addlInners
+          )
 
         val (polygons, _) = (dissolvedOuters ++ addlOuters)
           // sort by size (descending) to use rings as part of the largest available polygon
-          .sortWith(_.area > _.area)
+          .sortWith(_.getArea > _.getArea)
           // only use inners once if they're contained by multiple outer rings
           .foldLeft((Vector.empty[Polygon], dissolvedInners)) {
           case ((ps, is), outer) =>
-            val preparedOuter = outer.prepare
-            (ps :+ Polygon(outer.exterior, is.filter(inner => preparedOuter.contains(inner)).map(_.exterior)), is.filterNot(inner => preparedOuter.contains(inner)))
+            val preparedOuter = preparedFactory.create(outer)
+            (
+              ps :+ GTPolygon(
+                outer.getExteriorRing.getCoordinateSequence,
+                is
+                  .filter(inner => preparedOuter.contains(inner))
+                  .map(_.getExteriorRing.getCoordinateSequence)
+                ).jtsGeom,
+              is.filterNot(inner => preparedOuter.contains(inner))
+            )
         }
 
         Some(polygons match {
-          case Vector(p: Polygon) => p.toWKB(4326)
-          case ps => MultiPolygon(ps).toWKB(4326)
+          case Vector(p: Polygon) => p
+          case ps => factory.createMultiPolygon(ps.toArray)
         })
       } catch {
         case e@(_: AssemblyException | _: IllegalArgumentException | _: TopologyException) =>
@@ -556,36 +626,42 @@ package object osm {
     (_: Map[String, String]) ++ (_: Map[String, String])
   }
 
-
   // TODO this (and accompanying functions) doesn't belong here
-  def buildRoute(id: Long, version: Int, timestamp: Timestamp, types: Seq[Byte], roles: Seq[String], wkbs: Seq[Array[Byte]]): Option[Seq[(String, Array[Byte])]] = {
-    if (types.zip(wkbs).exists { case (t, g) => t == WayType && Option(g).isEmpty }) {
+  def buildRoute(
+    id: Long,
+    version: Int,
+    timestamp: Timestamp,
+    types: Seq[Byte],
+    roles: Seq[String],
+    geometries: Seq[Geometry]
+  ): Option[Seq[(String, Geometry)]] = {
+    if (types.zip(geometries).exists { case (t, g) => t == WayType && Option(g).isEmpty }) {
       // bail early if null values are present where they should exist (members w/ type=way)
       logger.debug(s"Incomplete relation: $id @ $version ($timestamp)")
       None
     } else {
-      val geoms = wkbs.map(Option(_).map(_.readWKB) match {
-        // polygons are not part of routes (unless they're platforms)
-        case Some(geom: Line) => geom.as[Line]
-        case _ => None
-      })
+      val geoms = geometries.map { geom =>
+        geom match {
+          // polygons are not part of routes (unless they're platforms)
+          case l: LineString => Some(l)
+          case _ => None
+        }
+      }
 
       try {
         Some(roles.zip(geoms)
           .filter(_._2.isDefined)
           .map(x => (x._1, x._2.get))
-          .groupBy {
-            case (role, _) => role
-          }
+          .groupBy { case (role, _) => role }
           .mapValues(_.map(_._2))
           .mapValues(connectSegments(_))
           .map { case (role, lines) =>
             lines match {
-              case Seq(line) => (role, line.toWKB(4326))
-              case _ => (role, MultiLine(lines.toArray).toWKB(4326))
+              case Seq(line) => (role, line)
+              case _ => (role, factory.createMultiLineString(lines.toArray))
             }
-          }
-          .toSeq)
+          }.toSeq
+        )
       } catch {
         case e@(_: AssemblyException | _: IllegalArgumentException | _: TopologyException) =>
           logger.warn(s"Could not reconstruct route relation $id @ $version ($timestamp): ${e.getMessage}")
```

src/main/scala/vectorpipe/osm/ExtractMultiPolygons.scala
```diff
@@ -0,0 +1,125 @@
...
+ *
+ * spark-submit \
+ *   --class osmesa.ExtractMultiPolygons \
+ *   ingest/target/scala-2.11/osmesa-ingest.jar \
+ *   --orc $HOME/data/rhode-island.orc \
+ *   --cache $HOME/.orc \
+ *   --out $HOME/data/rhode-island-geoms
+ */
+
+object ExtractMultiPolygons extends CommandApp(
```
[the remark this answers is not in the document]
jbouffard: That's a good point. I'll make that project now.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,756 @@
...
+    *
+    * Nodes and ways contain implicit timestamps that will be used to generate minor versions of geometry that they're
+    * associated with (each entry that exists within a changeset).
+    *
+    * @param _ways        DataFrame containing ways to reconstruct.
+    * @param _nodes       DataFrame containing nodes used to construct ways.
+    * @param _nodesToWays Optional lookup table.
+    * @return Way geometries.
+    */
+  def reconstructWayGeometries(_ways: DataFrame, _nodes: DataFrame, _nodesToWays: Option[DataFrame] = None)(implicit cache: Caching = Caching.none, cachePartitions: Option[Int] = None): DataFrame = {
```
mojodna: I initially added the `implicit Caching` options to facilitate debugging interim outputs; it probably makes sense to remove them now, esp. since they're not used by any of the construction methods.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
mojodna: Yeah, best to remove this until there's a clearer need for this in the future.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
mojodna: (`Caching.scala` can potentially go away entirely.)

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,676 @@
...
+
+  private val MultiPolygonTypes = Set("multipolygon", "boundary")
+
+  private val BooleanValues = Set("yes", "no", "true", "false", "1", "0")
+
+  private val TruthyValues = Set("yes", "true", "1")
+
+  private lazy val logger = Logger.getLogger(getClass)
+
+  private val factory = GeomFactory.factory
```
mojodna: Should any references to `factory` just point at `GeomFactory.factory`?

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,756 @@
...
+
+          val isArea = nds.head.getAs[Boolean]("isArea")
+          val geom = nds
+            .sortWith((a, b) => a.getAs[Int]("idx") < b.getAs[Int]("idx"))
+            .map { row =>
+              Seq(Option(row.get(row.fieldIndex("lon"))).map(_.asInstanceOf[Float]).getOrElse(Float.NaN),
+                  Option(row.get(row.fieldIndex("lat"))).map(_.asInstanceOf[Float]).getOrElse(Float.NaN))
+            } match {
+              // no coordinates provided
+              case coords if coords.isEmpty => Some(GeomFactory.factory.createLineString(Array[Coordinate]()))
```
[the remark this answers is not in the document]
jbouffard: It looks like it does:

```
scala> GeomFactory.factory.createLineString(Array[Coordinate]())
res0: com.vividsolutions.jts.geom.LineString = LINESTRING EMPTY
```

I don't think you could pass in an empty `Array` to a GT `Line`, though.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
mojodna: Awesome. `ReversedCoordinateSequence`, `PartialCoordinateSequence`, and `VirtualCoordinateSequence` **definitely** shouldn't be public, since they don't provide immutable guarantees and implement some functions fast and loose.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
jbouffard: Ah, okay. I wasn't sure if the `Caching` was doing something more involved. I will remove those then.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,756 @@
...
+      .where(
+        ('memberUpdated.isNull and 'memberValidUntil.isNull and 'geom.isNull) or // allow left outer join artifacts through
+          ('memberUpdated <= 'updated and 'updated < coalesce('memberValidUntil, current_timestamp)))
+      .drop('memberUpdated)
+      .drop('memberValidUntil)
+      .drop('ref)
+
+    implicit val encoder: Encoder[Row] = VersionedElementEncoder
+
+    // leverage partitioning (avoids repeated (de-)serialization of merged coordinate arrays)
```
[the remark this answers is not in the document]
jbouffard: Thanks for pointing that out. I'll remove it.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
jbouffard: Yeah, that'd probably be better.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,676 @@
...
+
+    override def setOrdinate(i: Int, ordinateIndex: Int, value: Double): Unit = {
+      val (sequence, index) = getSequence(i)
+
+      sequence.setOrdinate(index, ordinateIndex, value)
+    }
+
+    // TODO this should be invalidated after append (but it doesn't actually matter because all of the appending will
+    // occur ahead of time)
+    private lazy val coordinates: Array[Coordinate] = {
```
mojodna: This is one of those fast and loose spots where invalidating is skipped because we know `coordinates` will only be requested after the sequence has been fully assembled.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,676 @@
...
+            }
+        }, rings)
+    }
+  }
+
+  implicit val geometryFactory: GeometryFactory = new geom.GeometryFactory()
+
+  private def formRings(segments: GenTraversable[LineString])(implicit geometryFactory: GeometryFactory): GenTraversable[Polygon] =
+    formRings(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
+      .map(geometryFactory.createPolygon)
```
mojodna: Now that this is surfaced by GT, the `implicit val` and arg can go away in favor of accessing it directly.

```scala
  private def formRings(segments: GenTraversable[LineString]): GenTraversable[Polygon] =
    formRings(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
      .map(GeomFactory.factory.createPolygon)
```

This line was also the impetus for adding some additional `Polygon` constructors (https://github.com/locationtech/geotrellis/pull/2618/files#diff-2268e822bd244d311242d85a54f766d0R40), I think this is actually even simpler and doesn't require access to the `GeometryFactory`:

```scala
  private def formRings(segments: GenTraversable[LineString]): GenTraversable[Polygon] =
    formRings(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
      .map(Polygon(_))
```

mojodna: Though, right, these are JTS geometries, not GT ones, so the first is correct.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
```diff
@@ -0,0 +1,676 @@
...
+  private def connectSegments(segments: GenTraversable[LineString])(implicit geometryFactory: GeometryFactory): GenTraversable[LineString] =
+    connectSegments(segments.map(_.getCoordinateSequence).map(s => new VirtualCoordinateSequence(Seq(s))))
+      .map(geometryFactory.createLineString)
+
+  private def getHoles(poly: Polygon): List[CoordinateSequence] = {
+    for (i <- 0 until poly.getNumInteriorRing) yield
+      poly.getInteriorRingN(i).getCoordinateSequence
+  }.toList
+
+  private def dissolveRings(rings: GenTraversable[Polygon]): (Seq[Polygon], Seq[Polygon]) = {
```
mojodna: This no longer dissolves. (That's what the unary `union` in `MultiPolygon(rings.toArray).union.asMultiPolygon` did.)

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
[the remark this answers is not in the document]
mojodna: That's why then. Cool.

I'd still want to use `Array.empty[Coordinate]` here.

src/main/scala/vectorpipe/osm/internal/functions/package.scala
jbouffard: That's something I meant to comment on. Do you know how to make the JTS Geometries dissolve?

mojodna: Not off the top of my head.

jbouffard: Okay, I'll ask around and see if anyone knows

jbouffard: @mojodna So I think this does perform the dissolve; as it still dissolves all of the rings within the polygon.

mojodna: Actually, this is something that the unit tests in `MultiPolygonRelationReconstructionSpec` will check (there's at least 1 relation made up up multiple ways w/o rings), so perhaps it'd be best to add that and use it to confirm.

src/main/scala/vectorpipe/osm/internal/util/Caching.scala
```diff
@@ -6,7 +6,7 @@ import org.apache.spark.sql._
 import java.io.File
 
 
-trait Caching {
+trait Caching extends Serializable {
```
mojodna: Is this still used?

jbouffard: Just in the `OSMReader`, but I think that can probably be taken out

src/main/scala/vectorpipe/osm/internal/functions/package.scala
mojodna: I did a little poking; once you have a JTS `MultiPolygon`, call `.union()` on it and you should get either a `Polygon` or `MultiPolygon` back.

Something like this should work:

```scala
val mp = factory.createMultiPolygon(rings.toArray).union
val ps = for (i <- 0 until mp.getNumGeometries) yield mp.getGeometryN(i).asInstanceOf[Polygon]
(ps.map(_.getExteriorRing).map(x => factory.createPolygon(x.getCoordinateSequence)),
 ps.flatMap(getHoles).map(factory.createPolygon))
```

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,769 @@
...
+
+    @transient val idAndVersionByUpdated = Window.partitionBy('id, 'version).orderBy('updated)
+    @transient val idByUpdated = Window.partitionBy('id).orderBy('updated)
+
+    wayGeoms
+      // Assign `minorVersion` and rewrite `validUntil` to match
+      .withColumn("validUntil", lead('updated, 1) over idByUpdated)
+      .withColumn("minorVersion", (row_number over idAndVersionByUpdated) - 1)
+      .withColumn("geometryChanged", !((lag('geom, 1) over idByUpdated) <=> 'geom))
+      .join(ways.select('id, 'version, 'tags, 'visible, 'uid, 'user), Seq("id", "version"))
```
mojodna: This will only surface the correct `uid` and `user` for `minorVersion === 0` (it will assign authorship to the user who edited the **way** rather than the node that modified it), so it's probably better to join on changesets, since `changeset` will correspond to the changeset that triggered the minor version. You could flow `uid` and `user` through the explosion / geometry generation process, but that increases the shuffle substantially.

src/main/scala/vectorpipe/osm/internal/ProcessOSM.scala
```diff
@@ -0,0 +1,769 @@
...
+
+    val allRelationVersions = relationsByChangeset
+      // Union with raw relations to include those in the time line (if they weren't already triggered by geometry
+      // modifications at the same time)
+      .union(relations.select('changeset, 'id, 'version, 'timestamp as 'updated, 'uid, 'user))
+      // If a node, a way, and/or a relation were modified within the same changeset at different times, there will be
+      // multiple entries with different timestamps; this reduces them down to a single update per changeset.
+      .groupBy('changeset, 'id)
+      .agg(max('version).cast(IntegerType) as 'version, max('updated) as 'updated)
+      .join(relations.select('id, 'version, 'members, 'uid, 'user), Seq("id", "version"))
```
mojodna: This has the same issue as ways where authorship will be incorrect for minor versions.

There's another, not obvious at all (but alluded to in the comment below), problem that surfaced when generating geometries for the full planet history: in order to calculate `validUntil`, **all** relation rows need to be on the same partition for the window function to execute.

tl;dr: this function is very sensitive to the addition of new columns, especially columns that can be large, so it's worth seeing if data can be added after the explosion / construction stage.

jbouffard: Oh, I see. I didn't realize that there was a performance cost. I'll remove those extra columns then. Would adding the `uid` and `user` information after the dataframe is created be efficient? I was thinking about joining them by using the `id`s from the data source.

mojodna: Yes. Using `addUserMetadata` on the output of this function should do the trick.

PR #66, opened 2019-03-27
Fix checkpoint off-by-one

mojodna:
The `checkpoints` table contains sequence IDs that have been processed, so it should begin with the sequence immediately following rather than re-processing.


mojodna:
Ooh, it passes now.

PR #80, opened 2019-04-24
Add RELEASING.md and a changelog

### LINKED ISSUE #77: Add a project changelog
We've had good success with the human-curated [keepachangelog](https://keepachangelog.com/en/1.0.0/) format for projects we don't update super frequently.

### LINKED ISSUE #79: Add `RELEASING.md` to repo
Following the discussion in #78, it is worthwhile to codify the steps involved in making a release so that the commit log is readable, important changes are underscored in an accessible way (i.e., #77), that tags are maintained consistently, and binary artifacts are correctly delivered.  We should add a `RELEASING.md` file describing the SOP for releases to maintain continuity into the future.

CloudNiner:
# Overview

I had started this after the 1.0.0-RC3 release and so when the discussion happened in #78 and #79 got created, that provided an additional kick to clean up and PR my notes right away.

There's also a CHANGELOG in here, I'm open to keeping or discarding whichever changes you all feel are most appropriate at this time.

@mojodna I attempted to adjust my list to match the suggested workflow you provided in https://github.com/geotrellis/vectorpipe/pull/78#issuecomment-486362399, as I like that better than what I did for RC3 (I attached the RC3 tag to master after merging the PR). The git flow paradigm essentially follows the bullets you laid out.

Closes #79 
Closes #77

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,16 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+## [Unreleased]
+### Added
+- RELEASING.md - Instructions for releasing new versions of this project
+
+## [1.0.0-RC3] - 2019-04-24
+### Fixed
+- Mark all logger vals and some UDF vals as @transient lazy to avoid Spark serialization issues
+- Properly strip leading and trailing slashes from S3 URIs when exporting vector tiles
+
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -9,6 +9,7 @@ following the prompts after running `./sbt bintrayChangeCredentials`
 1. Create a new release branch from up-to-date master, e.g. `release/x.y.z`
 1. Bump version refs in README
 1. Bump version ref in `project/Version.scala`
+1. Review CHANGELOG.md. Move `[Unreleased]` header to empty section and replace with `[x.y.z]` header plus release date. 
 1. Commit these changes as a single commit, with the message "Release vx.y.z"
 1. Push branch and make a PR on GitHub
 1. Ensure CI succeeds
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,0 +1,19 @@
+# Releasing
+
+When releasing, ensure you have write access to the `azavea` 
+bintray organization. You can configure your credentials by
+following the prompts after running `./sbt bintrayChangeCredentials`
+
+## Publishing a release
+
+1. Create a new release branch from up-to-date master, e.g. `release/x.y.z`
+1. Bump version refs in README
+1. Bump version ref in `project/Version.scala`
+1. Commit these changes as a single commit, with the message "Release vx.y.z"
+1. Push branch and make a PR on GitHub
+1. Ensure CI succeeds
+1. Ensure there are no new commits on master, then merge. If there are new commits, rebase this branch on master and start over at step 2 if you wish to include them.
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,14 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
```
CloudNiner: > adheres to [Semantic Versioning]

Do we actually want this? If not, I'll remove this statement.

RELEASING.md
```diff
@@ -0,0 +1,20 @@
...
+1. Create a new release branch from up-to-date master, e.g. `release/x.y.z`
+1. Bump version refs in README
+1. Bump version ref in `project/Version.scala`
+1. Review CHANGELOG.md. Move `[Unreleased]` header to empty section and replace with `[x.y.z]` header plus release date. 
+1. Commit these changes as a single commit, with the message "Release vx.y.z"
+1. Push branch and make a PR on GitHub
+1. Ensure CI succeeds
+1. Tag the "Release vx.y.z" commit: `git tag -a vx.y.z -m "Release x.y.z"`
+1. Publish the tagged commit as a new release: `./sbt publish`
+1. Merge release branch
```
mojodna: What happens if another PR is merged (or commits pushed to `master`) before the release PR is merged?

mojodna:
> I attached the RC3 tag to master after merging the PR

This makes more sense to me, especially if there's a way to not create a merge commit (`git commit --ff`), since it means that the tip of the tree will get tagged vs. one of the parents of the merge commit (which may be older than the parent from `master`).

CHANGELOG.md
mojodna: I'm a fan, though what that means w/r/t dependencies _here_ should probably be spelled out (esp. since this isn't a standalone app).

jpolchlo: I think we're semver across our GT-related projects, so let's stay consistent.

RELEASING.md
jpolchlo: We take gun, we shoot guy who merge.</russian>

jpolchlo: Is it the right thing to move the actual merge up one or two spots?  The publish can happen later with no ill effects, and will shrink the window for bad merges/commits.  Although, for an organization that has few committers, I wonder how risky this is.

CHANGELOG.md
CloudNiner: Makes sense, I'll leave it in. 

> what that means w/r/t dependencies here should probably be spelled out

Sorry... I'm not totally sure what you mean. Upstream or downstream deps or both? Could you expand just a bit?

RELEASING.md
CloudNiner: Yeah. That's fine, I think that's how I originally had it, and is how I did the RC3 merge. I moved the merge item up to just after "Ensure CI succeeds". I also added a note to rebase on master and go back to step two if there are new commits on master.

CHANGELOG.md
mojodna: I was about what happens if we bump upstream deps.

Since deps are shared between this lib and the downstream project, changing a version here potentially has pile-on effects downstream.

CloudNiner: Ah. Yes. That makes sense. I would expect that any time we bump upstream deps, but the API of this lib remains backwards compatible, then we wouldn't release a new major version -- verified via tests and usage. If you're referring specifically to upstream GT libs, I'm less confident about that but by default I'd expect the same restrictions.  @jpolchlo do we have a precedent in the other GT libs?

jpolchlo: /shrug

Probably want input from @pomadchin or @echeipesh

PR #98, opened 2019-06-24
Checkpoint intermediate levels to reduce re-calculation

mojodna:
# Overview

With this change, runtime for `PipelineSpec` → "should generate multiple zoom levels" (which creates 2 intermediate levels) goes from ~70s to ~28s.

## Notes

`PipelineSpec` → "should generate multiple layers" will fail without #97.

/cc @echeipesh @jpolchlo


jpolchlo:
I validated this PR on https://github.com/geotrellis/geotrellis-osm-diff-demo.  I can confirm a roughly 40% improvement in performance:

Without checkpointing:

real	11m27.766s
user	28m50.020s
sys	1m5.398s

With checkpointing:

real	6m56.285s
user	14m38.264s
sys	0m36.820s

If we can figure out how to get this to pass Travis, then I'm ready to push the big green button on this.

Issue #51, opened 2017-12-24
Support MBtiles


kamicut:
[MBtiles](https://github.com/mapbox/mbtiles-spec) is a portable vector file format, storing tiles in an SQLite file. It is widely used in tools to [process and visualize vector tiles](https://github.com/mapbox/awesome-vector-tiles). Vectorpipe should support this output file format.

mojodna:
[`mbutil/patch`](https://github.com/mapbox/mbutil/blob/d495c0a737f4bce5fde8e707fe267674be39369a/patch) demonstrates how MBTiles archives can be merged together. This is the sort of thing that could be done when downloading + merging partition output from S3, similar to [`hadoop fs -getmerge`](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/FileSystemShell.html#getmerge), but content-aware.

PR #100, opened 2019-06-24
Return Dataset[BinaryTile] so generated MVTs can be stored in a single output

mojodna:
# Overview

This is in addition to (optionally) writing individual MVTs out. This will save on object management and time taken to `PUT` individual tiles.

The single output contained generated MVTs can be subsequently processed to extract specific tiles or to write to an alternate data sink (e.g. MBTiles (#51)).

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Support for `nds` in augmented diff GeoJSON (matching
     [`osm-replication-streams@^0.7.0`](https://github.com/mojodna/osm-replication-streams/tree/v0.7.0)
     output)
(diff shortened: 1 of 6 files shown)


jpolchlo:
I approve of this concept.  Would be good for benchmarking purposes as well to cut out the costly write operation.  Would love to see if we can come up with a way to generate an MBTiles file out of one of these `Dataset[BinaryTile]` guys.

jpolchlo:
Oh, but, we should hold off on deploying this until after 1.0.0

jpolchlo:
This (I think) has been obviated by #140.  Reopen if that's not true.

mojodna:
👋 Not quite (as far as I can imagine). I can imagine using `finalize` to output a single file(set) per zoom, but returning `Unit` from the end of the pipeline makes it difficult (impossible?) to accumulate (ideally lazily and optionally) all of the tiles for all of the zooms into a single data frame (and potentially re-group tiles from multiple zooms together).

PR #102, opened 2019-06-24
Filter out uninteresting tags

mojodna:
# Overview

Uninteresting tags are those that have little-to-no rendering relevance and shouldn't be considered when promoting nodes to points. The tag (and prefix) list was generated by querying for the top 200 single-use tags (i.e. the only tag applied to an element). `fixme`, `note`, and `source` are omitted, as they can be useful for validation and tracking.

## Checklist


mojodna:
After applying these filters to the 20190610 history dump, the number of generated point geometries goes from 557,657,578 vs 262,662,521.

PR #116, opened 2019-10-15
Upgrade to GT3.0 codebase

### LINKED ISSUE #115: Upgrade to GeoTrellis 3.0-SNAPSHOT
The API is not expected to change for the RC at this point, so upgrade to 3.0-SNAPSHOT is in order and required by some ongoing work.

moradology:
# Overview

This PR upgrades to GT 3.0. There are a few significant changes that deserve special attention in this regard:
1. Spray -> Circe
2. Geotrellis Geometries -> JTS Geometries
3. AWS SDK V1 -> AWS SDK V2
4. Overriding dependencies to avoid test failures due to `jackson` version 2.9 colliding with Spark's `jackson` version

Closes #115

### DIFF
--- a/project/Version.scala
+++ b/project/Version.scala
@@ -1,6 +1,6 @@
 object Version {
   val awscala = "0.8.1"
-  val vectorpipe = "1.2.0-SNAPSHOT"
+  val vectorpipe = "2.0.0-SNAPSHOT"
   val scala = "2.11.12"
   val geotrellis = "3.0.0-SNAPSHOT"
   val geomesa = "2.2.1"
--- a/project/Version.scala
+++ b/project/Version.scala
@@ -2,7 +2,7 @@ object Version {
   val awscala = "0.8.1"
   val vectorpipe = "1.2.0-SNAPSHOT"
   val scala = "2.11.12"
-  val geotrellis = "2.2.0"
+  val geotrellis = "3.0.0-SNAPSHOT"
   val geomesa = "2.2.1"
   val decline = "0.6.1"
   val cats = "1.6.0"
--- a/src/main/scala/vectorpipe/vectortile/Pipeline.scala
+++ b/src/main/scala/vectorpipe/vectortile/Pipeline.scala
@@ -1,10 +1,9 @@
 package vectorpipe.vectortile
 
-import geotrellis.vector.{Feature, Geometry}
 import geotrellis.layer._
+import geotrellis.vector._
 
 import org.apache.spark.sql.{DataFrame, Row}
-import org.locationtech.jts.{geom => jts}
 
 /**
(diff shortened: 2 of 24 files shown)


project/Version.scala
```diff
@@ -2,7 +2,7 @@ object Version {
   val awscala = "0.8.1"
   val vectorpipe = "1.2.0-SNAPSHOT"
```
jpolchlo: I want to suggest a version number bump here, either to a new major version number or to apply a "-GT3" suffix.  I'm not sure I know what semver would have to say in this case.

CloudNiner: Looks like an externally visible breaking change would be the swap from GT geom types to JTS? Along with some import changes, such as `store` and `layer`? 

If we're semvering and there's breaking changes, that's a major version bump. If we need to do bugfix releases supporting gt 2 those can continue to be minor version or bugfix changes on VP 1.0.

So I'd definitely agree with a version bump, and additionally suggest major bump to VP 2.0.0

jpolchlo:
The true test of this would be if we can get OSMesa working with this new version without problem.  I don't see any changes that are likely to block that, and we don't strictly need this confirmation, since the pressure to update versions isn't coming from that side.  I would say we could/should defer that test, and call this PR provisionally OK.

src/main/scala/vectorpipe/vectortile/Pipeline.scala
```diff
@@ -148,5 +148,5 @@ trait Pipeline {
    * See [[geotrellis.vectortile.Value]] for details.
    */
   def pack(row: Row, zoom: Int): VectorTileFeature[Geometry] =
-    Feature(Geometry(row.getAs[jts.Geometry](geometryColumn)), Map.empty)
+    Feature(row.getAs[jts.Geometry](geometryColumn), Map.empty)
```
jpolchlo: Prefer here to not import JTS, import all of `vector` and drop these `jts.` prefixes.  This comment will surely apply at other points in this PR.  `jts.geom` should basically not be imported unless you're working with coordinates.

jpolchlo:
So, for this PR, we should prefer `import geotrellis.vector._` to importing JTS, I think.  We surely _can_ import items from JTS, but in the interest of making sure GT's convenience extensions are available, the GT import is better.

CloudNiner:
@jpolchlo 

> So, for this PR, we should prefer import geotrellis.vector._ to importing JTS

Done. See rebased commit https://github.com/geotrellis/vectorpipe/pull/116/commits/1d0f3bafe0f99612f504fbd1b2368e9fa5d013ae

jpolchlo:
Pending Travis approval, this looks good to me.

PR #120, opened 2019-11-25
Use CircleCI to publish to Sonatype

### LINKED ISSUE #117: Switch deployment to SonaType
In the previous months we have been switching our deployments forgo BinTray and switch ti publishing directly to SonaType (ex: https://github.com/geotrellis/geotrellis-contrib/pull/236)

This gives us SNAPSHOT releases and enables better CI workflow on the projects.
This issue is to switch vectorpipe project similar release structure.

This will probably mean that the group ID for this project will switch from `org.azavea` to `org.azavea.geotrellis`:

https://github.com/geotrellis/vectorpipe/blob/a1ecfc6fc77afd8bb0f2141abb5f532f57b57167/build.sbt#L8

I think this is fine and good, that consistent with the rest of the projects published from `github.com/geotrellis` and reflects the reality that VectorPipe is pretty much part of GeoTrellis ecosystem of tools.

jpolchlo:
# Overview

Another step in the "eradicate Travis and Bintray" project undertaken by many GT ecosystem projects.

## Checklist

Closes #117


jpolchlo:
Tagging @geotrellis/operations to get some clarity on this

project/Version.scala
```diff
@@ -1,8 +1,9 @@
 object Version {
   val awscala = "0.8.1"
   val vectorpipe = "2.0.0-SNAPSHOT"
```
echeipesh: unused now since we pull version from git ?

jpolchlo: Unless it's valuable for local publishing for testing purposes?

echeipesh: Its completely unused now:

https://github.com/geotrellis/vectorpipe/pull/120/files#diff-fdc3abdfd754eeb24090dbd90aeec2ceR8-R14

The  `gitDescribedVersion` includes a SHA for the snapshot release `git describe --tags`

jpolchlo: :guardsman: Aye, cap'n

.circleci/config.yml
```diff
@@ -0,0 +1,118 @@
...
+      #      - "openjdk8-scala2.12.8"
+      #    filters:
+      #      tags:
+      #        only:
+      #          - /^(.*)$/
+      #      branches:
+      #        only:
+      #          - develop
+      #          - /release\/.*/
+      #          - /hotfix\/.*/
```
jpolchlo: This comment block is temporary until Geomesa publishes a spark 2.4-compatible release.  We can cross-compile to scala 2.12 at that point and uncomment the above.

build.sbt
```diff
@@ -77,17 +82,61 @@ lazy val commonSettings = Seq(
...
+  publishTo := sonatypePublishTo.value
+)
+
+lazy val credentialSettings = Seq(
+  credentials += Credentials(
+    "GnuPG Key ID",
+    "gpg",
+    System.getenv().get("GPG_KEY_ID"),
+    "ignored"
+  ),
```
echeipesh: @hectcastro Kind of late to the party here, but what is the mechanism that prevents the gpg from asking for the passphrase during the publish? Is this somehow side effect of `gpg --batch` command in `cibuild` ?

jpolchlo: Possibly unrelated, but `publishLocal` gives
```
[warn] GPG reported an error. Artifacts won't be signed.
```
Also, there is an NPE through log4j raised just before that message pops up.

FFT?

rbreslow: Yup, this is related to the line you saw in the CircleCI YAML.

The way we get around GPG asking for a passphrase is super simple: we simply do not password-protect the signing key. This is OK, because the signing key is a subkey that is only used in CI and is very easy to invalidate.

Better docs on this workflow will be dropping soon in the Operations repo.

See: https://github.com/jodersky/sbt-gpg#securely-publishing-in-travis-and-other-ci

echeipesh:
This looks like its working and done, going to merge and try to cut a tagged releases with this.

build.sbt
echeipesh: @jpolchlo It was pulling in an env var that was not defined, thus producing `null` value, I added a commit to flatten credentials from vars that don't exist.

PR #122, opened 2019-12-11
Move VectorPipe examples from azavea/osmesa

CloudNiner:
# Overview

Add stream processing examples with `Source`, `Changeset`, etc to `vectorpipe.examples`. These examples lived in [azavea/osmesa](https://github.com/azavea/osmesa) but were moved here as part of https://github.com/azavea/osmesa/pull/177 since they really demonstrate VectorPipe functionality.

The old examples all constructed a consistent, default SparkSession. That was moved here as well, and is now provided as `VectorPipe.defaultSparkSessionWithJTS`.

I elected not to split VP into subprojects at this time, the list of examples doesn't feel large or comprehensive enough yet to warrant it. Instead I opened https://github.com/geotrellis/vectorpipe/issues/123 for the future.

## Testing Instructions

You should be able to get these examples to run by following the docstring comment found in each. 

## Checklist

### DIFF
--- a/src/main/scala/vectorpipe/VectorPipe.scala
+++ b/src/main/scala/vectorpipe/VectorPipe.scala
@@ -5,15 +5,19 @@ import vectorpipe.vectortile.export._
 
 import geotrellis.proj4.{CRS, LatLng, WebMercator}
 import geotrellis.layer._
+import geotrellis.spark.store.kryo.KryoRegistrator
 import geotrellis.vector._
 import geotrellis.vectortile._
 
+import org.apache.spark.SparkConf
 import org.apache.spark.rdd.RDD
+import org.apache.spark.serializer.KryoSerializer
 import org.apache.spark.sql._
 import org.apache.spark.sql.catalyst.expressions.GenericRowWithSchema
 import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.StringType
 import org.apache.spark.storage.StorageLevel
+import org.locationtech.geomesa.spark.jts.SparkSessionWithJTS
 
 object VectorPipe {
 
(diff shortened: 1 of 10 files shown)


src/main/scala/vectorpipe/VectorPipe.scala
```diff
@@ -151,7 +155,34 @@ object VectorPipe {
...
+      .set("spark.ui.showConsoleProgress", "true")
+      .set("spark.serializer", classOf[KryoSerializer].getName)
+      .set("spark.kryo.registrator", classOf[KryoRegistrator].getName)
+
+    SparkSession.builder
+      .config(conf)
+      .enableHiveSupport
+      .getOrCreate
+      .withJTS
+  }
```
jpolchlo: Ooo.  Nice.  Me likey.

src/test/scala/vectorpipe/TestEnvironment.scala
```diff
@@ -30,7 +30,7 @@ object TestEnvironment {
 trait TestEnvironment extends BeforeAndAfterAll { self: Suite with BeforeAndAfterAll =>
   implicit val ss: SparkSession = SparkSession.builder
     .master("local[*]")
-    .appName("OSMesa Test")
+    .appName("VectorPipe Test")
```
jpolchlo: Good catch

CloudNiner:
>  Isn't it nice when you talk about doing a thing, and you manage to do exactly that thing with no surprises? I wonder what that's like.

idontalways

src/main/scala/vectorpipe/VectorPipe.scala
CloudNiner: Whew.

I'll be curious to see if this gets used anywhere else.

PR #135, opened 2020-02-04
Add call to ensureCompressedMembers in preprocessRelations

### LINKED ISSUE #134: Broken joins between relation members using $"member.type" as String and lit(WayType) as Byte
When I run `OSM.toGeometry(osmFeatures)` it eventually fails with:
```
Exception: java.lang.String cannot be cast to java.lang.Byte
	at scala.runtime.BoxesRunTime.unboxToByte(BoxesRunTime.java:93)
	at vectorpipe.internal.package$$anonfun$15$$anonfun$16.apply(package.scala:526)
```
So digging into the code, I see that the relation member `type` column is expected to be a Byte as one of the OSM type constants defined in `vectorpipe/internal/package.scala`
https://github.com/geotrellis/vectorpipe/blob/685df3a02d28c5f39fc8182ca6f1897885a0b8b4/src/main/scala/vectorpipe/internal/package.scala#L526

But as the exception shows, the `type` column is actually a String and I believe that is because `getRelationMembers()` selects `$"member.type"` from the input relations as the `type` which is the raw String representation (`"node"`, `"way"`, or `"relation"`).
https://github.com/geotrellis/vectorpipe/blob/685df3a02d28c5f39fc8182ca6f1897885a0b8b4/src/main/scala/vectorpipe/internal/package.scala#L332

Later in `getRelationMembers()` it tries to join those relation members with the geometries which contain the Byte representation of the `type`, and so I think this join probably doesn't work.
https://github.com/geotrellis/vectorpipe/blob/685df3a02d28c5f39fc8182ca6f1897885a0b8b4/src/main/scala/vectorpipe/internal/package.scala#L349
https://github.com/geotrellis/vectorpipe/blob/685df3a02d28c5f39fc8182ca6f1897885a0b8b4/src/main/scala/vectorpipe/internal/package.scala#L351

Then the `getRelationMembers()` selects the String `$"member.type"` and returns that DataFrame which eventually results in the ClassCastException when attempting to read the `type` column as a Byte.
https://github.com/geotrellis/vectorpipe/blob/685df3a02d28c5f39fc8182ca6f1897885a0b8b4/src/main/scala/vectorpipe/internal/package.scala#L383

JonMcPherson:
# Overview

The README claims the String to Byte conversion or relation member `type` is "handled for you" in the top-level OSM object, but that is not the case.  So this PR adds a call to `ensureCompressedMembers` within `preprocessRelations` to the filtered `"relation"` records which will have non-empty `members` arrays.
Note: I figured it made sense to do this conversion when pre-processing input relations rather than only in the top-level OSM object (as the README claims).

## Notes

See Issue https://github.com/geotrellis/vectorpipe/issues/134

## Testing Instructions

The counts printed by `ProcessOSMTest` did not change after this change, and I removed the call to `ensureCompressedMembers` on the input test data since that would now be redundant.

## Checklist

Closes #134

### DIFF
--- a/src/main/scala/vectorpipe/internal/package.scala
+++ b/src/main/scala/vectorpipe/internal/package.scala
@@ -136,10 +136,11 @@ package object internal {
     } else {
       @transient val idByUpdated = Window.partitionBy('id).orderBy('version)
 
+      val relations = ensureCompressedMembers(history.where('type === "relation"))
+
       // when an element has been deleted, it doesn't include any tags; use a window function to retrieve the last tags
       // present and use those
-      history
-        .where('type === "relation")
+      relations
         .repartition('id)
         .select(
           'id,
--- a/src/test/scala/vectorpipe/ProcessOSMTest.scala
+++ b/src/test/scala/vectorpipe/ProcessOSMTest.scala
@@ -2,7 +2,6 @@ package vectorpipe
 
 import org.scalatest._
 import vectorpipe.{internal => ProcessOSM}
-import vectorpipe.functions.osm.ensureCompressedMembers
 
 class ProcessOSMTest extends FunSpec with TestEnvironment with Matchers {
   val orcFile = getClass.getResource("/isle-of-man-latest.osm.orc").getPath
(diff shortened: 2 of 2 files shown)


JonMcPherson:
@jpolchlo no problem! I wanted to learn more of the Vectorpipe internals anyway since I just discovered this project (and GeoTrellis) after finishing the POC of a very similar Spark application for converting OSM data (in Parquet format) to vector tiles with clipped geometries using JTS and GeoTools...  I had to deal with all of the same OSM data nuances that are handled by Vectorpipe, however Vectorpipe (and GeoTrellis) are of course **much** more robust and feature complete, so I am currently working on migrating to Vectorpipe + GeoTrellis (and it's awesome so far!)

Yea I think more test for this behavior would be good.
I'll keep playing around with this awesome project and will post any other issues or PRs as I learn more.

jpolchlo:
Thanks for your contribution.  Feel free to file issues as you come up with questions!

PR #138, opened 2020-06-08
Receive GPG key while publishing artifacts

rbreslow:
## Overview

We've [added](http://keyserver.ubuntu.com/pks/lookup?search=0x52D6103D31A87CB6&fingerprint=on&op=index) a new signature to the GeoTrellis public key to extend the validity of the signing subkey. 

This change set receives the latest copy of the public key at the start of the `cipublish` build stage, allowing us to "renew" the key in the future without needing to update the key pair stored in CircleCI secrets.

Additionally, this PR cleans up some things that were left behind in #120. See comments for more details around that.

### Checklist

- [ ] Description of PR is in an appropriate section of the CHANGELOG and grouped with similar changes if possible

## Testing Instructions

TBD.

Connects https://github.com/azavea/operations/issues/446


.circleci/config.yml
```diff
@@ -40,23 +40,12 @@ aliases:
...
       - image: circleci/openjdk:8-jdk
     environment:
       SCALA_VERSION: 2.11.12
 
-  - &openjdk8-scala2_12_8-environment
-    docker:
-      - image: circleci/openjdk:8-jdk
-    environment:
-      SCALA_VERSION: 2.12.8
-
```
rbreslow, line 59: Since we're not building on Scala 2.12.8, we don't need these.

.circleci/config.yml
```diff
@@ -76,27 +65,7 @@ workflows:
                 - /^(.*)$/
             branches:
               only:
-                - develop
-                - /release\/.*/
-                - /hotfix\/.*/
-                - feature/circleci
```
rbreslow, line 82: Since this repository is not git flow-enabled, it would never build on any of these branches. If we want to publish snapshots, then we could add `master` to this list, otherwise it should only publish tagged commits.

rbreslow:
@CloudNiner Could you leave a comment describing the changes that you've made relating to dependencies? I'm curious to know what happened, and think it'd be valuable for anyone who takes a look at this in the future.

build.sbt
```diff
@@ -46,10 +46,10 @@ lazy val commonSettings = Seq(
     Resolver.bintrayRepo("lonelyplanet", "maven"),
     Resolver.bintrayRepo("kwark", "maven"), // Required for Slick 3.1.1.2, see https://github.com/azavea/raster-foundry/pull/1576
     Resolver.bintrayRepo("bkirwi", "maven"), // Required for `decline` dependency
-    "locationtech-releases" at "https://repo.locationtech.org/content/repositories/releases/",
-    "locationtech-snapshots" at "https://repo.locationtech.org/content/repositories/snapshots/",
+    "eclipse-releases" at "https://repo.eclipse.org/content/groups/releases",
+    "eclipse-snapshots" at "https://repo.eclipse.org/content/groups/snapshots",
     "geosolutions" at "http://maven.geo-solutions.it/",
-    "boundless" at "https://repo.boundlessgeo.com/main/",
+    "osgeo-releases" at "https://repo.osgeo.org/repository/release/",
```
pomadchin, line 52: 1. locationtech repos migrated to eclipse
2. boundless repos migrated to osgeo

build.sbt
```diff
@@ -46,10 +46,10 @@ lazy val commonSettings = Seq(
...
     "osgeo" at "http://download.osgeo.org/webdav/geotools/",
```
pomadchin, line 53: We can also remove this repo, since it is dead now as well.

project/Dependencies.scala
```diff
@@ -21,7 +21,7 @@ object Dependencies {
   val cats           = "org.typelevel"               %% "cats-core"                   % Version.cats
   val scalactic      = "org.scalactic"               %% "scalactic"                   % Version.scalactic
   val scalatest      = "org.scalatest"               %%  "scalatest"                  % Version.scalatest % "test"
-  val jaiCore        = "javax.media"                 % "jai_core"                     % "1.1.3"
+  val jaiCore        = "javax.media"                 % "jai_core"                     % "1.1.3" from "https://repo.osgeo.org/repository/release/javax/media/jai_core/1.1.3/jai_core-1.1.3.jar"
```
pomadchin, line 24: This is the change to make sbt happy with the jai_core dep. In the new GT releases it would not be neccesary to add jai_core like this.

However, this fix does not guarantee that all builds would sucesfully download it. To guarantee it we need to exclude jai_core from all deps to avoid accidential bad resolutions. 

I experimented with it and cam to a conclusion that we can exclude all jai_core deps from all the 3d party deps and in this case we guarantee the resolution odering. [That is the fix for GeoTrellis](https://github.com/locationtech/geotrellis/blob/master/project/Settings.scala#L278).

CloudNiner: Gotcha, so I still need to add the exclude in your link to the build.sbt for vectorpipe.

pomadchin: @CloudNiner well we can live without it for now I guess

CloudNiner:
@rbreslow do @pomadchin 's inline comments answer your questions? Short version is this build.sbt defined some external repos that were moved to other locations, and jai core is a frustrating dependency where we applied geotrellis's fix here.

.circleci/config.yml
```diff
@@ -76,27 +65,7 @@ workflows:
...
-      #    filters:
-      #      tags:
-      #        only:
-      #          - /^(.*)$/
-      #      branches:
-      #        only:
-      #          - develop
-      #          - /release\/.*/
-      #          - /hotfix\/.*/
+                - jrb/receive-gpg-key
```
ColeKettler, line 68: 📌 Tacking a reminder here to remove this branch filter prior to merge

PR #150, opened 2021-06-10
Enable CircleCI for v1 line

jpolchlo:
# Overview

Now that bintray is sunsetting, we have no means to publish v1 artifacts.  This PR aims to bring in the CircleCI/sonatype changes on master to these old versions.


pomadchin (approved):
🚀

